PR #30, opened 2025-04-06
Add Tectonic Pokedex

AlphaKretin:
There's a lot of work to be done on the Pokedex before it's ready, but I want to unify these two projects at this stage in development so that I can share resources like data and images


AlphaKretin:
There's more work to be done, but it's at a competent enough draft stage for me to want to push it to live... as soon as I figure out why the build is breaking on Vercel

PR #57, opened 2025-04-16
Load data from Tectonic repo

### LINKED ISSUE #56: Automatic, inbuilt PBS parsing
This could also include an automatic versioning system for team exports.
Unfortunately this would require converting all of the static data fetching to async methods, which is a pain to get working in React.
Instead I could perhaps rework my manual update pipeline to use the code from this and work all in one step.
This would also mean far fewer calls to fetch the data.

AlphaKretin:
Close #56 

The following data still needs to be updated to the new system/format:
- [ ] Encounters
- [x] Forms
- [ ] Trainers
- [ ] Trainer types

The following data needs to be loaded instead of hard-coded:
- [x] Types
- [x] Tribes
- [x] Type Chart
- [ ] Version (currently not implemented at all)

Other to-dos:
- [ ] Refactor systems to remove redundant handling (e.g. evolutions)
- [x] Handle lack of nulls in new system
- [ ] Fix matchup display with new type chart


AlphaKretin:
Given that `techo` is a dev branch, I'm comfortable merging this in its current state and making separate issues for the remaining to-dos

Issue #65, opened 2025-04-16
Refactor evolution handling to use data added by new PBS parsing


AlphaKretin:

Cincidial:
@AlphaKretin if you haven't started on this one I can it pick up?

AlphaKretin:
@Cincidial Absolutely, go ahead!

AlphaKretin:
forgot to have the pr merge close this! 🎉

PR #71, opened 2025-04-17
#65 Evolution Refactor

Cincidial:
Needs some polish still, but can probably be done later so this isn't blocking. 
- Sizing, mainly lines like scyther, but could use improvements elsewhere


AlphaKretin:
I'll take a look and at least clean up the build errors

src/app/pokedex/components/PokemonModal.tsx
```diff
@@ -249,11 +244,10 @@ const PokemonModal: React.FC<PokemonModalProps> = ({ pokemon: mon, onClose }) =>
...
                                     onClick={() => handleTabChange(tab)}
-                                    className={`px-4 py-2 text-sm font-medium ${
-                                        activeTab === tab
-                                            ? "text-blue-600 border-b-2 border-blue-600 dark:text-blue-400 dark:border-blue-400"
-                                            : "text-gray-500 hover:text-gray-700 dark:text-gray-300 dark:hover:text-gray-100"
-                                    }`}
+                                    className={`px-4 py-2 text-sm font-medium ${activeTab === tab
+                                        ? "text-blue-600 border-b-2 border-blue-600 dark:text-blue-400 dark:border-blue-400"
+                                        : "text-gray-500 hover:text-gray-700 dark:text-gray-300 dark:hover:text-gray-100"
+                                        }`}
```
Cincidial, line 250: Note for myself, looks like my vscode auto formats this to be different on save.

AlphaKretin: I'll try and copy my VSCode settings locally so they're part of the repo

Cincidial:
> I'll take a look and at least clean up the build errors

Cool cool, no worries though if you are working on other things. I'll continue working through this tomorrow.

Cincidial:
Also, just tagging the related issue #65

src/app/data/util.ts
```diff
@@ -11,3 +11,84 @@ export function isNull(o: Pokemon | Move | Trainer | Ability | Item | undefined)
...
+
+    ascendParents(fn: (node: NTreeNode<T>) => void) {
+        let node: NTreeNode<T> | null = this;
+        while (node != null) {
+            fn(node)
+            node = node.parent;
+        }
+    }
+
+    asBreadthFist(): NTreeNode<T>[][] {
```
Cincidial, line 82: Typo to fix

src/app/pokedex/components/PokemonModal.tsx
```diff
@@ -413,44 +406,39 @@ const PokemonModal: React.FC<PokemonModalProps> = ({ pokemon: mon, onClose }) =>
...
-                                                        {describeEvoMethod(prevo)}
-                                                    </li>
-                                                ))}
-                                            </ul>
-                                        </div>
-                                    ) : (
-                                        <p className="text-gray-600 dark:text-gray-300">No previous evolutions.</p>
-                                    )}
+                                    {currentPokemon.evolutionTree.isLeaf()
+                                        ? (<p className="text-gray-600 dark:text-gray-300">No further evolutions.</p>)
```
Cincidial, line 410: Forgot to change this text.

AlphaKretin:
image
Issue: first-stage pokemon are marked as a previous evolution for themselves

src/app/data/util.ts
```diff
@@ -71,11 +71,12 @@ export class NTreeNode<T> {
         return null;
     }
 
+    // recursion pattern modified to avoid aliasing "this" to local variable
+    // hopefully still works as expected
```
Cincidial, line 75: Oh no yeah this is totally the way to do it, dunno what I was doing haha

src/app/data/signatures.ts
```diff
@@ -81,7 +81,7 @@ function getMoveLearnableGroups(): MoveCounts {
     });
 
     Object.values(pokemon).forEach((speciesData) => {
-        if (speciesData.getEvos().length > 0) return;
+        if (speciesData.isFinalEvo()) return;
```
AlphaKretin, line 84: This (and its cousin 40 lines up) should be negated, the way Tectonic assigns signatures is to *only* final evos

Cincidial:
@AlphaKretin - aside from anything the comes up with review this should be okay to bring in. Some polish needed still, but can be done later so it's not blocking.

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "prettier.tabWidth": 4,
+    "prettier.printWidth": 120,
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
```
Cincidial, line 4: I added this line so it should only apply to this project, I think?
Otherwise if set in VSCode user settings it will apply to other projects

AlphaKretin: Yep this is a good call

Cincidial:
For sure, I just kinda tossed some stuff in utility and was hoping you had a preferred place haha.

Yeah I'm still thinking of how to fix trees like that. That reavor column would essentially need a fake element above it to push it down. 
Alternatively, each evolution branch can instead be rendered on its own - so in this case there would be 2 rows both showing scyther.

AlphaKretin:
Actually I'm gonna merge first and then make the changes I'm thinking of. 🚀

AlphaKretin:
Something I had on my mind that I forgot to say or test was if the NTreeNode in LoadedPokemon would mess with the data saving, but I just gave it a test now and it seems not! Phew.

Cincidial:
@AlphaKretin - I tried having it be saved so it didn't need to be generated, but yeah that caused cyclic dependency.

So it just saves with null and is built for the user when their scripts get loaded

PR #75, opened 2025-04-18
Add option to fetch data from Tectonic development branch

### LINKED ISSUE #72: Allow optionally getting and using data from Tectonic dev branch
Would be useful for developers and contributors, would allow earlier testing of the versioning in the team builder, and would allow us to prepare in advance for the pokemon.txt format changes I know are coming.
Possibly have as a second hosted instance, rather than an option on the public website. This would avoid having to write a bunch of branches into the website code, and if desirable might allow keeping the dev version restricted to devs.

AlphaKretin:
Close #72 
To load data from dev instead of main, run the `npm run data` command with the extra `dev` parameter, i.e. `npm run data dev`.

### DIFF
--- a/src/app/data/loading/loadData.ts
+++ b/src/app/data/loading/loadData.ts
@@ -7,7 +7,7 @@ import { parseForms } from "./forms";
 import { parseItems } from "./items";
 import { parseMoves } from "./moves";
 import { parsePokemon, parsePokemonLegacy, propagatePokemonData } from "./pokemon";
-import { parseTribes } from "./tribes";
+import { parseTribes, parseTribesLegacy } from "./tribes";
 import { buildTypeChart, TypeChart } from "./typeChart";
 import { parsePokemonTypes } from "./types";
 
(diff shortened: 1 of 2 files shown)


AlphaKretin:
The backend data should all be updated with this, just need to decide how to use it 
- Allow choosing between data sets during runtime (most convenient for users, hardest to implement for us)
- Save multiple data sets and allow choosing between them on boot (good for archival, still non-trivial to implement)
- Have one data set loaded and leave it up to the host to change it when desired (no changes needed to implement, more effort for the host to maintain)

AlphaKretin:
Confirmed team import/export works across versions. *May* want to add explicit handling for move/ability legality changes, but the current symptom isn't the biggest deal.

For reference, team codes for my "Charmer" team seen on the old teambuilder spreadsheet:
`3.2.1!AMMB1QB2AAAB4QEMARkCvw==!ASwAqgBZAAAAGwQkA/gAGQ==!AnwA+wB+AAAB4QNrAPMBUA==!AiQBbwBfAAADlgJQA6cBlg==!ATEAbP//AAAA0ALmAJYBLA==!A5sBoQBVAAABLgQaAgwAGw==`

`dev!ANcB4gB3AAABwQD3AQMClA==!AZoApwBaAAAAFgQhA/P//w==!AvsA/QB/AAABwQNRAN8BOg==!ApIBeABgAAADgAInA48Bew==!AaEAa///AAAAwQK6AIsBFg==!AEgBqwBWAAABGAQVAeYAFg==`

AlphaKretin:
I suppose since one of the possible modes proposed above is currently functional, I can merge this and then make an issue to discuss if we change it

Issue #73, opened 2025-04-17
Improve appearance of Scyther evolution line


AlphaKretin:
![image](https://github.com/user-attachments/assets/9058de4a-431f-439b-8257-ca4e4ca96634)
> Yeah I'm still thinking of how to fix trees like that. That reavor column would essentially need a fake element above it to push it down.
Alternatively, each evolution branch can instead be rendered on its own - so in this case there would be 2 rows both showing scyther.

AlphaKretin:
Rendering multiple Scythers wouldn't be the worst option, the current solution does look a little janky with the two arrows not quite actually coming from it.

Cincidial:
I'll pick this issue up and see if there's a way to get things lookin' nice.

AlphaKretin:
forgor to make the pr merge close this again

PR #85, opened 2025-04-18
#73 Evolution Polish

Cincidial:
Made each evolution branch it's own row to account for cases like scyther and snorunt, who have paths with different depths. In general, most pokemon only have one path so it looks the same as before, for more specific cases though here's some screen shots:

image
image
image


Cincidial:
Actually, wait I see a lot of changes in main, just gunna sync and see if anything breaks lol

AlphaKretin:
Vercel seems to be confused by the changes to the build process happening in the middle of the PR but it all looks good on my end so hopefully it'll work when it's merged to main. Thanks for this!

AlphaKretin:
yea we're good

Issue #86, opened 2025-04-18
Apply abilities to STAB matchup table


Cincidial:
Was thinking this could be a cool idea similar to how the defensive matchups table works. 
For example, something with breakthrough would instead list a stab electric vs ground as regular instead of immune. 

Not sure if makes sense in the modal as it's only stab then, maybe better just left for the team builder. So, close this if it doesn't make sense to you.

AlphaKretin:
I think it's not completely unreasonable to have it affect the Pokedex modal, it's useful to know that Pikachu with Break Through has a different offensive profile to, say, Voltorb, even though the offensive matchup table is only every an estimate. 

It does make *more* sense in the team builder of course, it's a higher priority there if it's not already implemented.

AlphaKretin:
*narrator voice* it was already implemented there

Issue #62, opened 2025-04-16
Don't display multiple rows in matchup modal if abilities don't affect it


AlphaKretin:

Cincidial:
I'll grab this one

PR #94, opened 2025-04-19
#62 & #86: Update def and stab match up table to handle having or not having an ability that affects them

Cincidial:
image
image
image
image
image

UI is now slightly changed from before as I've split the type chart into 2 rows, similar to how PokemonDB does it. 
This way the user doesn't have to scroll to see more.

### DIFF
--- a/src/app/data/types/PokemonType.ts
+++ b/src/app/data/types/PokemonType.ts
@@ -1,3 +1,5 @@
+import { LoadedType } from "../loading/types";
+
 export class PokemonType {
     id: string;
     index: number;
@@ -7,22 +9,14 @@ export class PokemonType {
     immunities: string;
     isRealType: boolean;
 
-    constructor(
-        id: string,
-        index: number,
-        name: string,
-        weaknesses: string,
-        resistances: string,
-        immunities: string,
-        isRealType: boolean
-    ) {
-        this.id = id;
-        this.index = index;
-        this.name = name;
-        this.weaknesses = weaknesses;
-        this.resistances = resistances;
-        this.immunities = immunities;
-        this.isRealType = isRealType;
+    constructor(type: LoadedType) {
+        this.id = type.key;
+        this.index = type.index;
+        this.name = type.name;
+        this.weaknesses = type.weaknesses;
+        this.resistances = type.resistances;
+        this.immunities = type.immunities;
+        this.isRealType = type.isRealType;
     }
 
     getShortName(): string {
--- a/src/app/data/types/PokemonType.ts
+++ b/src/app/data/types/PokemonType.ts
@@ -1,4 +1,4 @@
-export interface PokemonType {
+export class PokemonType {
     id: string;
     index: number;
     name: string;
@@ -6,4 +6,26 @@ export interface PokemonType {
     resistances: string;
     immunities: string;
     isRealType: boolean;
+
+    constructor(
+        id: string,
+        index: number,
+        name: string,
+        weaknesses: string,
+        resistances: string,
+        immunities: string,
+        isRealType: boolean
+    ) {
+        this.id = id;
+        this.index = index;
+        this.name = name;
+        this.weaknesses = weaknesses;
+        this.resistances = resistances;
+        this.immunities = immunities;
+        this.isRealType = isRealType;
+    }
+
+    getShortName(): string {
+        return this.name.substring(0, 3).toUpperCase();
+    }
 }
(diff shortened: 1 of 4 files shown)


Cincidial:
Handles both #86 and #62

src/app/data/types/PokemonType.ts
```diff
@@ -1,9 +1,31 @@
...
+        this.id = id;
+        this.index = index;
+        this.name = name;
+        this.weaknesses = weaknesses;
+        this.resistances = resistances;
+        this.immunities = immunities;
+        this.isRealType = isRealType;
+    }
+
+    getShortName(): string {
```
Cincidial, line 28: Originally was going to do this with short names to make the table smaller, but it looked pretty not good haha. Left it like this in case it can be reused later, but if this messes up something can revert back to an interface.

src/app/pokedex/components/PokemonModal.tsx
```diff
@@ -257,83 +304,97 @@ const PokemonModal: React.FC<PokemonModalProps> = ({ allMons, pokemon: mon, hand
...
-                                                    {realTypes.map((t) => {
-                                                        const mult = calcTypeMatchup(
-                                                            { type: t },
-                                                            {
-                                                                type1: currentPokemon.getType1(currentForm),
-                                                                type2: currentPokemon.getType2(currentForm),
-                                                                ability: a,
-                                                            }
+                                    {realTypesSlices.map((slice, index) => (
+                                        <table key={index} className="table-fixed w-[99%] mx-auto mb-5">
```
Cincidial, line 329: w-[99%] because full width was causing the right most border to not be displayed because??? I'm assuming the scroll bar shadow or something?

AlphaKretin:
I like the table being split! I saw it in Techo but didn't put in the effort to copy it ajskdf;lja
Gastly doesn't have a toggle button on the atk matchups despite having Break Through. It does have a button for Levitate on def matchups, and Pikachu does have a button for *its* break through on atk matchups, so idk exactly what's going on.

src/app/data/types/PokemonType.ts
AlphaKretin: I'm happy leaving it as is!

Cincidial:
> I like the table being split! I saw it in Techo but didn't put in the effort to copy it ajskdf;lja Gastly doesn't have a toggle button on the atk matchups despite having Break Through. It does have a button for Levitate on def matchups, and Pikachu does have a button for _its_ break through on atk matchups, so idk exactly what's going on.

Lmao what the heck. I just kinda assumed all would work after checking some haha

Will take a look!

AlphaKretin:
> > I like the table being split! I saw it in Techo but didn't put in the effort to copy it ajskdf;lja Gastly doesn't have a toggle button on the atk matchups despite having Break Through. It does have a button for Levitate on def matchups, and Pikachu does have a button for _its_ break through on atk matchups, so idk exactly what's going on.
> 
> Lmao what the heck. I just kinda assumed all would work after checking some haha
> 
> Will take a look!

I only took a quick glance at the code, but I think it might only be checking if one of the abilities changes the matchups, assuming the other would be neutral? Could be wrong in that analysis though.

Cincidial:
Ah so I think I know what's going on. 
So to do the check the code:
- assumes false for the first ability, as if there is only 1 ability we don't need it to be swappable
- for the second ability it compares each type matchup calc value with the previous abilities value. If one of them is different then we need the button to swap because the 2 abilities result in different calcs.

Did it this way so it's super general and doesn't actually need any per ability checks other than the regular type calc.

However in the case of stab it tries both types and takes the max. Which means for ghastly
- Breakthrough changes 0 to 1 for poison against steel, but ghost against steel is already 1. So effectively there's no difference to the type chart
- likewise breakthrough changes 0 to 1 for ghost against normal, but poison against normal is already 1 so there's no difference.

So attack matchups for mons with 2 types likely won't be showing the button because the other type will cover for the immunity. Effectively giving the same type chart with or without the ability.

Can maybe change to be 2 seperate type charts, 1 for each of the moms types rather than a combined one?

Thats quite a bit of UI space and info, so might be a bit confusing

AlphaKretin:
Oh, right! If the ability has no effect, then it doesn't need a toggle, yeah, I was thinking about the Ghost side of things and forgot that Poison already granted neutral coverage there. In that case, this should all be working, my bad!

Cincidial:
No worries lol, when you mentioned it I was also very confused too haha. I totally thinking it should appear to. But yeah, there's an option to make it showable.

Also, maybe more work, could introduce a "move type selection" there and then it applies the ability with the move type selected. Probably a bit excessive for the pokedex modal tho lol, probably fits better on the team builder

AlphaKretin:
apologies for the delay in merging this, I'm just finishing up something on my end before I can focus on it

Cincidial:
> apologies for the delay in merging this, I'm just finishing up something on my end before I can focus on it

No worries, you're actually lightning fast. 
At my work everything moves slow, average PR time is a month, sometimes 2. There's a couple at 8 months right now lmao.

PR #95, opened 2025-04-19
Add spec document for team code format

### LINKED ISSUE #89: Write formal specification for team code format

AlphaKretin:
Close #89 

Leaving this open for a bit to get feedback from others.


AlphaKretin:
The example team code provided is currently in the outdated 1-item format. It is being used as a placeholder and should be replaced as soon as the implementation is updated to support 2 items.

AlphaKretin:
Seeking feedback both on the writing of the spec document, and the spec of the format itself, making sure I'm not missing anything that I won't regret leaving out later. I *believe* this is all the information necessary to reconstruct a Pokemon as it appear on an in-game team, Player or Trainer, in a mechanical sense - leaving out aesthetic details such as traits, nicknames, and hue.

AlphaKretin:
I've been reminded that some items, like the Crystal Veil, have additional parameters that should most likely also be encoded. How to do this efficiently when the parameters can be arbitrary is a tricky issue, but before I think about that I need to implement it at all in the code

AlphaKretin:
The codes are getting longer, and it's full of increasingly many mostly-undefined values. I'm tempted to change the format so that items and item types are at the end, and it can cut off undefined values from the end to trim space.

AlphaKretin:
OK, *now* I can't think of anything else to add to the spec

Cincidial:
> @Cincidial curious if you have any thoughts on this

I got some ideas. It's not gunna result and quick to understand spec, you will save quite a few bytes though. Just trade offs, so depends on what you want. I'll get a lil table going (I love thinking about this stuff)

Cincidial:
image
- Version: Major.Minor.Patch.dev where dev, forming the first 2 bytes
- Pokemon + Item 1 Type form bytes 3 and 4 of the first int32. 
-- Pokemon with 11 bits allows 2048 total and 5 bits for item type gives 32 possibilities
- The style points + level at 5 bits per style point and 7 bits for level (meeting all max values) make up 1 int32
- 2 groups of 2 moves + an item each form an int32
-- May be able to limit move indices to what the pokemon can learn, but didn't here. Can maybe save 2 bytes per pokemon this was. 11 bits for a move gives 2048 possibilities, right now there are 1123
- Finally item 2, forms and ability are in the last 2 bytes
-- Forms: Most here is 27 for unknown so 6 bits giving 64 values should be plenty
-- Abiliity: Only 1 bit required because we know the pokemon, so just index it's abilities.

In total
- 16 bytes for pokemon + 2 for version
- 98 total bytes for a full team

For what's in the document currently (from what I understand)
- 34 bytes per pokemon + 6 for version + 6 to insert `!` characters
- 216 total bytes for a full team

In both cases the base64 conversion will increase byte size by 25% because of the 8->6 byte reduction per character.
Thus max expected character length would go from 270 -> 123

There's the chance to add in the trimming as you've described to further reduce, but with added complexity during processing

AlphaKretin:
I was considering an approach like this where I used less space for values where we could afford it, but I was having trouble implementing it correctly and so went with the simpler "all UInt16" approach. Would you be willing to give a try at implementing this scheme? No worries if not, like you said it's a trade-off and I think I'd be happy sticking with what I've got, but I am intrigued by the possibility of the more compact option.

Cincidial:
Oh most assuredly I'll give it go! Honestly love doing the bitwise stuff.
Will check it out in the morning

AlphaKretin:
Needs to be rewritten if we go with #121

AlphaKretin:
As Tectonic implementation has been going well enough, the decision has been made to go with the scheme from #121, so this needs to be rewritten when that's finished.

AlphaKretin:
@Cincidial Got something here as a start but it could use a look over from the person who actually put most of the work into defining the format

team-code-format-specification.md
```diff
@@ -0,0 +1,268 @@
...
+   d. Map index values back to their corresponding entities using version-specific mapping tables
+   e. Construct a Pokémon object with the decoded properties
+
+## 7. Version Mapping
+
+The binary format uses compact indices that are mapped to and from actual game data through version-specific mapping tables. These mappings convert between binary indices and game object identifiers for:
+
+- Pokémon species
+- Moves (per Pokémon)
+- Items
```
Cincidial, line 165: Probably mention that these are held items only (pocket 5)

Cincidial (commented):
Dang, this looks amazing. I love me some good documentation.
Just one comment, didn't really see anything else. I'll check it over again tomorrow and see if anything jumps out, but if you merge before then no worries as well

AlphaKretin:
When 3.3.0 releases for real I want to update the code to not be from the dev branch, so that it's a public, stable example, but we can edit that later without it being an actual change to the spec, so it shouldn't block merging this branch.

That said, since it's info only and not code or anything, there is no rush, I'm happy to wait for another review.[

AlphaKretin:
Should probably have a note about which versions are expected to be supported, just realised that got forgotten in the rewrite. I previously said it's up to implementation, but it probably couldn't hurt to suggest that versions 3.2.4 and up are supported (with dev versions probably being optional).

It's probably kind of hypothetical as to what other applications would actually use the format, but y'know.

Cincidial:
Did another review, nothing stood out. Looks good!

AlphaKretin:
Yippee!

PR #96, opened 2025-04-19
Add proper validation for two-item abilities

### LINKED ISSUE #44: Support multiple held items in team builder
Should only be enabled when ability allows

AlphaKretin:
Close #44


AlphaKretin:
Notably, two-item abilities do have a flag which we could in theory check, but we need the hardcoded data for what that check *is* anyway, so checking that map is sufficient.

PR #101, opened 2025-04-19
Move shared functionality for Pokemon calculation into class

### LINKED ISSUE #100: Better reuse code/logic between builder and calculator
There's a lot of shared functionality and logic between the team builder and the damage calculator - inputting Pokemon details, calculating the effect of selected abilities/items/etc on type matchups, etc... It would be for the best if they shared the code for this where possible 

In particular, it's very likely that the damage calculator could make use of the PokemonCard data structure currently used for the team builder, with the addition of stat steps

AlphaKretin:
Close #100 

- [ ] Refactor the handling of custom move parameters to extract the input component from the move class and have it work more like the Crystal Veil item
- [x] Parse item types from PBS and include them when loading TrainerPokemon into the calculator


AlphaKretin:
I think this is mostly all there? There was some stuff I wanted to smooth out first but I got sidetracked by the other stuff going on and don't remember most of it. 

Right, I wanted to see if I could change the whole MoveCustomVar situation to work more like the type does for items., since the solution of having React Nodes in the classes is uh... janky at best, it feels like. 

Also I haven't actually *tested* this yet and I suspect it will look pretty awkward with the card nested inside the column. But y'know. Something something function first visuals later.

So yeah I'll come back to this for that. After I get some shut eye (or fail to again)

AlphaKretin:
merged main in first so that the build hopefully stops failing and i can stop seeing the red X everywhere. *then* going to bed.

AlphaKretin:
The card surprisingly doesn't look *too* bad, but mostly by accident because the backgrounds blur together. It's still unnecessarily squished and there's a barely perceivable border. I'll try making the PokemonCard component only contain the internals.

AlphaKretin:
Also, todo: load item type data from trainers.pbs

AlphaKretin:
Gonna make that refactor a separate PR because I'm very unsure about the approach I'm taking. With that, I think everything else is good

Issue #38, opened 2025-04-06
Fix row hover effect on multi-type pokemon


AlphaKretin:

Cincidial:
I'm assuming this is the on hover color change in the big list?
I can pick this one up

AlphaKretin:
Yep yep. I think the issue is that the gradient overwrites the solid colour bg setting. I've tried setting the gradient to off when hovered but I probably didn't do it right, I'm flying by the seat of my pants with this stuff.

PR #103, opened 2025-04-19
#38: Fix hovering & some minor polish to evolution display

Cincidial:
Did some polish to the evolution display. I noticed that for the beedrill line the width get really cramped. 
So changed to a flex div instead of table. Also ended up improving the other tree types as well (I think)

For the hovering stuff... I cannot believe that took like 5 hours to figure out lmao.
Idk if it does this for you, but I cannot change global.css locally otherwise the site loses the majority of the tailwind css.
Even just changing the font in body causes that. 
For some reason git stashing then popping fixes it, but if I just revert the file and save it it's still broken.

Anyways, changed the on hover to invert the color scheme of the row and text but leave the img and buttons alone via a fake css class used for selection


package.json
```diff
@@ -18,14 +18,14 @@
     },
     "devDependencies": {
         "@eslint/eslintrc": "^3",
-        "@tailwindcss/postcss": "^4.0.15",
+        "@tailwindcss/postcss": "^4.1.4",
```
Cincidial, line 21: Happened while I was trying to fix tailwind stuff, can revert if you don't want to update

src/components/PokemonGradient.css
Cincidial, line 1: idk where you want css that isn't in global, I literally could not get this compile in global.css though. No clue why. My partner also watched and tried to help figure out why, we're both stumped though.

src/components/colours.ts
```diff
@@ -1,194 +1,16 @@
 import { PokemonType } from "@/app/data/types/PokemonType";
 import { Pokemon } from "../app/data/types/Pokemon";
+import "./PokemonGradient.css";
 
-export function getTypeBadgeColourClass(type: PokemonType) {
```
Cincidial, line 4: Converted to paramertized function. Drops the default case, meaning if there is a type that's not supported it would have an issue. Though if that happens I suspect you'd want to update to support that anyways

AlphaKretin:
image
This is, uh, not working for me, presumably because of whatever weirdness is going on with the CSS - it's not loading the *global* one at all even without having modified it. This is probably where not knowing what I'm doing in this area is coming back to bite me. I assume there's something up with how I've got everything configured, I'll look into it.

AlphaKretin:
Also, this is an idle aside aimed at the way the CI works, not at you, but it's annoying me that these dev preview builds keep failing. Even when I do authorise them the PR never gets the memo that it's started working.

package.json
AlphaKretin: No worries

AlphaKretin:
...maybe it would have actually worked this time, except that all dev builds are breaking because of a recent commit to Tectonic, lol.

AlphaKretin:
Oh, hang on, I see at least one of the issues going on here. This is something I ran into myself, and is why I had a slightly angry (previously very angry) comment in colours.ts. You can't dynamically construct Tailwind class names like that, Tailwind will only preload classes it finds written out in full in your source files. This is very silly, yes.

AlphaKretin:
> Oh, hang on, I see at least one of the issues going on here. This is something I ran into myself, and is why I had a slightly angry (previously very angry) comment in colours.ts. You can't dynamically construct Tailwind class names like that, Tailwind will only preload classes it finds written out in full in your source files. This is very silly, yes.

This being the case, I'm not quite sure how you got it working at all 😅. Reverting the function paramaterisation (but maintaining the inclusion of the class name), I've got gradients back, but the inversion doesn't work.

AlphaKretin:
I found a tailwind class to handle the inversion, but I'm struggling to make it *not* invert the badges and images, which I imagine might be why you used the approach you did

Cincidial:
> image This is, uh, not working for me, presumably because of whatever weirdness is going on with the CSS - it's not loading the _global_ one at all even without having modified it. This is probably where not knowing what I'm doing in this area is coming back to bite me. I assume there's something up with how I've got everything configured, I'll look into it.

Yeah this exactly how it looked for me without everything broken lol

Cincidial:
> I found a tailwind class to handle the inversion, but I'm struggling to make it _not_ invert the badges and images, which I imagine might be why you used the approach you did

Yup that's what I had run into

Cincidial:
> Oh, hang on, I see at least one of the issues going on here. This is something I ran into myself, and is why I had a slightly angry (previously very angry) comment in colours.ts. You can't dynamically construct Tailwind class names like that, Tailwind will only preload classes it finds written out in full in your source files. This is very silly, yes.

Oh did not know that, yeah that's rather silly

AlphaKretin:
OK yeah, on my end, with both local testing and the Vercel preview, I'm not seeing any inversion, with the classes set up as they are. Could this be a browser difference? I'm using Firefox.

AlphaKretin:
Wait hang on ugh I removed the CSS import while testing and didn't put it back

Cincidial:
I'm on Firefox based as well (floorp)

Cincidial:
I found a SO post saying you can do the dynamic CSS as long as even just comment in some file mentions the tailwind class.

So that could be an option lol, a fille just filled with comments

AlphaKretin:
OK every issue but the tailwind class thing was my fault for messing around oops. Turns out when you delete the code someone added, it stops working!! Can confirm invert-on-hover is working for me now.

AlphaKretin:
> I found a SO post saying you can do the dynamic CSS as long as even just comment in some file mentions the tailwind class.
> 
> So that could be an option lol, a fille just filled with comments

lol this makes sense. god. i might actually try that. 

As far as the hover thing goes: I don't really want to mess with it any further since it's working and seems to be volatile, so I'm probably going to leave the CSS structure as it is! Maybe at some point we'll do a big cleanup pass of everything and it can be tidied up then.

Cincidial:
Also something called safe listing 
https://tailwindcss.com/docs/detecting-classes-in-source-files#safelisting-specific-utilities

AlphaKretin:
Force pushed to undo my changes since I mostly just made a mess of things. Will now try adding comments to fix the missing tailwind classes

AlphaKretin:
> Also something called safe listing https://tailwindcss.com/docs/detecting-classes-in-source-files#safelisting-specific-utilities

From what I recall when I tried to look into this, safe-listing is considered removed/deprecated in newer versions of tailwind.

Cincidial:
I can take a look later as well.
Glad to know what was causing my global.caa issues tho, was super odd and was worried that file was some black magic somewhere haha

Cincidial:
> > Also something called safe listing https://tailwindcss.com/docs/detecting-classes-in-source-files#safelisting-specific-utilities
> 
> From what I recall when I tried to look into this, safe-listing is considered removed/deprecated in newer versions of tailwind.

Dang, that docs is 4.1 so I was hoping it would still be available

AlphaKretin:
> > > Also something called safe listing https://tailwindcss.com/docs/detecting-classes-in-source-files#safelisting-specific-utilities
> > 
> > 
> > From what I recall when I tried to look into this, safe-listing is considered removed/deprecated in newer versions of tailwind.
> 
> Dang, that docs is 4.1 so I was hoping it would still be available

Huh, actually, yeah, this isn't the same thing I saw being called removed before. I'll give it a try

AlphaKretin:
I tried adding a safelist as per the docs:
```css
@import "tailwindcss";
@source inline("{to,from,bg}-{fire,water,steel,rock,psychic,poison,normal,ice,ground,grass,ghost,flying,fighting,fairy,electric,dragon,dark,bug,mutant}{-bg,}");

:root {
```
But I get the error "`@source` paths must be quoted.", which is proving *very* difficult to Google (as was a previous error I *did* figure out that was a missing semicolon).

AlphaKretin:
The comment solution worked. 🤷‍♀️

AlphaKretin:
And with the tailwind issues out of the way, this works fine in globals.css (at least on my end)

AlphaKretin:
Sorry for all the chaos on this one!

Cincidial:
No problem! Honestly I had no clue what tailwind was doing there as well. Glad it's sorted

PR #106, opened 2025-04-20
Refactor handling of custom move parameters

AlphaKretin:
Externalise input and state to the component for rendering move details in the calculator

Look at all those unknowns! Is there a better way to do this?


AlphaKretin:
I'm quite a bit happier with this, I think. Less clutter in unrelated files. Of course now we just need to resolve all these conflicts...

PR #121, opened 2025-04-21
#95: Compacting the Team Build URL

Cincidial:
I'd say WIP in the sense as there's some cleanup needed of the javascript code, but the URL code generated worked (at least using the same team generated by the example currently in place)

Here's the new code:
```
BEMBwLDEjKUpSgQgwD8FABEtjKUpSgJQNBMCgtp9jKUpSgSgkAsDwKoljKUpShKwAAUtA4JhMozFKQokcABmDQMEk5yMpSlKAhAEDQA=
```

Vs the old code
```
3.2.1!AMMB1QAAAEYACgAKAAoACgAKAeEBDAEZAr8Adg==!ASwAqgAAAEYACgAKAAoACgAKABsEJAP4ABkAWQ==!AnwA+wAAAEYACgAKAAoACgAKAeEDawDzAVAAfg==!AiQBbwAAAEYACgAKAAoACgAKA5YCUAOnAZYAXwBh!ATEAbAAAAEYACgAKAAoACgAKANAC5gCWASwAe///AA0=!A5sBoQAAAEYACgAKAAoACgAKAS4EGgIMABsAVQ==
```

If you want to proceed with this the specs doc would need to be updated as well. 
Here's a screenshot of the new format. There were some changes for making data fall off as needed. 
image

WIP for cleanup

- [x] - Use utils where possible
- [x] - Check for typos and stuff, there's one in one of the comments
- [x] - Cleanup usage by calling code & remove uneeded classes / interfaces
- [x] - Implement TODO for version mapping
- [x] - Make the item pocket 5 for held items thing a util somewhere
- [x] - UI change to force the ability to be set to the first one by default
- [ ] - Add docs about limits (number of items supported by each section)
- [x] Handle base64 safe encoding


AlphaKretin:
> **tectonic-tools-dev** 	❌ Failed ([Inspect](https://vercel.com/lunas-projects-95729d7b/tectonic-tools-dev/J41TTSCzPRuboou4mP6vTJ4SCBPM)) 			Apr 21, 2025 1:39am

The dev build failure seems to just be the same unused import issue the old one had, building on an outdated version because of the weirdness with authorisation. i.e., not a concern, should hopefully rerun next time a change is made (and as above there's an issue to resolve)

AlphaKretin:
Another error in testing: I used the example team, and then set Lilligant's ability to null (which by necessity also removed the second item). Upon import, I got `Error: data.moves[moveIndex] is undefined` - notably moves, not abilities, which implies that it's handling null values poorly and/or that it's misaligning things on import.

Cincidial:
Oh I forgot to mention as part of the WIP stuff.
This code is assuming a UI change to force auto select the first ability instead of null. 
Since a pokemon cannot not have an ability (I think?) there's no reason to support nulls

AlphaKretin:
Oh, yeah that's not unreasonable. The idea was to have null mean "undecided" basically, but yeah if it's useful for compacting the format we can assume one. Let me try nulling something else lol.

AlphaKretin:
Nulled Volcarona's item this time and got the same `Error: offset is outside the bounds of the DataView` error as all-nulls. 

`BEMBwLDEjKUpSgQgwD8FABEtjKUpSgJQNBMCgtp9jKUpSv/wkAsDwKoljKUpShKwAAUtA4JhMoylKUokcABmDQMEk5yMpSlKAhAEDQA=`

Cincidial:
I'll get it fixed up with the rest of the null stuff though. 
That item one is what probably failed the default all null. Was hoping having -1 translate to an undefined value would result in skipping stuff gracefully, but guess not lmao.

Cincidial:
> As with whether to proceed with it overall, I mean. If it works! It's definitely more compact, and it's impressive work I'd hate to put to waste. My main concerns are about my own ability to understand it, to write the spec doc and reimplement in Tectonic (export-only probably, in ruby). I definitely get the bigger picture, I'd just be worried about getting into the weeds with all the details of the individual bit operations being correct.

I can help write the spec doc if you'd like. 
As far as actually using it though, don't worry about wasted work. I for some reason really enjoying writing stuff like this and don't get many opportunities to do so! Basically the journey was worth it anyways (:

I also haven't done ruby before, but could probably help get this in there if needed / wanted. No worries on that though / if there's a desire to not have a bunch of bitwise stuff like this in there. Definitely understandable.

AlphaKretin:
> > As with whether to proceed with it overall, I mean. If it works! It's definitely more compact, and it's impressive work I'd hate to put to waste. My main concerns are about my own ability to understand it, to write the spec doc and reimplement in Tectonic (export-only probably, in ruby). I definitely get the bigger picture, I'd just be worried about getting into the weeds with all the details of the individual bit operations being correct.
> 
> I can help write the spec doc if you'd like. As far as actually using it though, don't worry about wasted work. I for some reason really enjoying writing stuff like this and don't get many opportunities to do so! Basically the journey was worth it anyways (:
> 
> I also haven't done ruby before, but could probably help get this in there if needed / wanted. No worries on that though / if there's a desire to not have a bunch of bitwise stuff like this in there. Definitely understandable.

That's understandable! In that case I think what I'll to do decide is see about implementing it in Tectonic, whether or not I get overwhelmed, and go with it if I can pull it off, otherwise stick to the simpler scheme.

Cincidial:
Note, codes generated previously for testing won't work still. Part of the issue was the encoder

AlphaKretin:
Full example team: `BEMBwLDEjKUpSgQgwD8FABEtjKUpSgJQNBMCgtp9jKUpSgSgkAsDwKoljKUpShKwAAUtA4JhMoylKUokcABmDQMEk5yMpSlKAhAEDQA=`

Volcarona null item: `BEMBwLDEjKUpSgQgwD8FABEtjKUpSgJQNBMCgtp9jKUpSg/wkAsDwKoljKUpShKwAAUtA4JhMoylKUokcABmDQMEk5yMpSlKAhAEDQA=`

All nulls: `BEP///gAjKUpSg////////gAjKUpSg////////gAjKUpSg////////gAjKUpSg////////gAjKUpSg////////gAjKUpSg////8A`

All import as expected! Thanks for this. It might sit around for a bit while I try to get around to trying it out in Tectonic.

Cincidial:
Anytime! If you run into anymore cases lemme know / or have questions on anything.
I will note just to bring attention to it: the right shifting by the decoder to get the final value via `>>>` is necessary over it's brother `>>`. The triple shifts 0's while the double shifts 1's from the left. You'll want 0's shifted in to get the correct final bits. 

I'll leave cleaning up the code till you decide on moving forward with it or not. But, just remember not to merge if you do (: 
Some stuff is a lil messy (I think mostly the calling into and out of these methods now / use of utils and what not).

Cincidial:
Ah shoot one other thing - JS functions used here default to `big endian`. Not sure what ruby defaults to.

AlphaKretin:
I started investigating implementing this into Tectonic, and noticed something - it only supports choosing one ItemType, no matter what. It's not actually tied to the items, it's just a single property on the trainer Pokemon that's checked if the user has a relevant item in either slot. We should update how we handle item types here to reflect that, which would also allows us to remove the second item type from this schema.

AlphaKretin:
> @AlphaKretin got a question on how forms work. Since I saw unkown had the mots forms at 27 I set the bits so we can have up to 64. But, how does form selection actually work for the team builder? I don't see all the unkown variants, and probably don't need them cause it's really not necessary lol. Just wondering if you know what the max forms select-able is?

Unown *should* have its forms selectable, I don't have any logic filtering them by mechanical differentiation or anything (though maybe we could and should). I'll investigate (maybe tomorrow though, Australia time). The most number of forms is likely either Unown or Pikachu, with all of its Cosplay and Hat variants (not accessible in Tectonic but technically in the data).

AlphaKretin:
I'm realising the reason dev builds keep failing is because I need to authorise each one individually, it doesn't carry over for the PR from the first 💀
You applied for some sort of access for the main one right? That's why those ones are working? I can't find the setting to add you on my end, but if there's a way you could do the same for the dev deployment that would simplify things.

AlphaKretin:
Continuing to look into Tectonic implementation, though technically I haven't gotten to the part that would vary depending on the approach yet, this seems like the most logical place to document the process and spitball about barriers.

In this case, I'm trying to construct the PBS indices using in-game methods, but
```rb
species = GameData::Species.keys.find_index(mon.species)
echoln(species)```
Gives me 1427 for Jellicent, instead of 713. I have a vague notion that all forms are entries in the species table, so it makes sense that this would throw this off.

However, while writing this I've rubber ducked and realised that the Pokedex number *should* be the index number + 1, at least on dev with my Pokedex reorganisation, and since this code is only going to be in Tectonic on dev/future versions, that heuristic should apply.

**Comment by AlphaKretin** [comment]
image
Got something working for Moves, and hopefully other things too

**Comment by AlphaKretin** [comment]
As per the mention above, I've made a start on the Tectonic implementation, but I'm running into issues and could use a second pair of eyes

**Comment by Cincidial** [comment]
> I'm realising the reason dev builds keep failing is because I need to authorise each one individually, it doesn't carry over for the PR from the first 💀 You applied for some sort of access for the main one right? That's why those ones are working? I can't find the setting to add you on my end, but if there's a way you could do the same for the dev deployment that would simplify things.

Did the access thing again, not sure what the setting is. But, yeah it seems like it's per PR lol

**Comment by AlphaKretin** [comment]
> Did the access thing again, not sure what the setting is. But, yeah it seems like it's per PR lol

Ugh, I got the request but I can only accept one request to one deployment as a free user. We'll have to deal.

**Comment by AlphaKretin** [comment]
> Gives me 1427 for Jellicent, instead of 713. I have a vague notion that all forms are entries in the species table, so it makes sense that this would throw this off.

Oh right, for the record, the actual issue here is that `GameData::whatever.keys` is a mix of numbers and symbols, which is why for moves and items I have to go out of the way to filter the list down first. We can apply that approach to Pokemon too if this estimate ever proves wrong.

**Comment by AlphaKretin** [comment]
As discussed in https://github.com/xeuorux/Pokemon-Tectonic/pull/136, we may have to change the order that `allMoves` returns its moves, as per `recalculate_learnable_moves` in `Species.rb`

**Comment by AlphaKretin** [comment]
Tectonic implementation is not *done* but the barrier is not related to the specific encoding scheme, so let's make it official that we're going with this one.

**Comment by Cincidial** [comment]
> If we're indexing abilities and moves locally to the pokemon, rather than globally, we shouldn't need to track those indices in `versions.json` anymore, right? That could cut down on its size, which is notable since it could grow over time as new patches come out. We can also apply the held item filtering at that level instead of later, if we're not already.

Correct, same with abilities. 
However, if a pokemon's moveset changes then the indices will be off. 
So if versioning is wanted it'll need to track the ability and move data in pokemon.json

**Comment by AlphaKretin** [comment]
> Correct, same with abilities. However, if a pokemon's moveset changes then the indices will be off. So if versioning is wanted it'll need to track the ability and move data in pokemon.json

Ah, good point. Maybe less necessary for abilities, since the index is only ever gonna be 0 or 1, and if a Pokemon's abilities change between versions it's natural enough to have the ability update on import, like it would for an update to the game itself. For moves though, yeah, keeping versions makes sense.

**Comment by AlphaKretin** [comment]
image
How did I manage to break Pokemon indexing...?

**Comment by AlphaKretin** [comment]
If I export the example team from the website now, I get `BEMYgdDEjKUpSgQhDCcbgfEtjKUpSgJRnG0RQsp9jKUpSgSggA4TBKoljKUpShKwsEgtAMFxMoylKUokcJxeDSMEM5yMpSlKAhIMjQA=`. It seems like the very first bit has changed, despite me not changing the way the `dexNum` is stored at all. I must be doing something wrong with the moves that's rippling out?

**Comment by AlphaKretin** [comment]
OK so let's look at a specific example to figure things out. Espeon's dex number is 216 on dev, 196 on main. Teddiursa is 319 on dev, 216 on main. So there we go. ...Did I just have the main data loaded instead of the dev data? I could have sworn it should have been the dev data. The items lined up correctly... but then they probably haven't changed a ton.

**Comment by AlphaKretin** [comment]
image
OK, yeah, that was just a skill issue on my part. Now. Delcatty and Aggron are loading their moves correctly - great! Others seem to not. However, this may be the fault of #131.

**Comment by AlphaKretin** [comment]
image
OK! I think it's a lot closer to correct now, it's just the order that linemoves are being propagated.

**Comment by AlphaKretin** [comment]
> OK! I think it's a lot closer to correct now, it's just the order that linemoves are being propagated.

OK that's not the *exact* issue, switching that around just made things worse. Notably, Volcarona uses tutormoves and is correct. I think it's just going to be down to subtleties of internal move ordering. Which might be a bit of a pain to dig into. I might try working on forms instead for now and leave the moves for later/a more careful eye.

**Comment by AlphaKretin** [comment]
My test case for forms has been lycanroc, and the *form* part seems to work (team code: `gGMaxqtwjKUpSm/xsFcAAg==`), but the ability is undefined. Possibly because it only has one? If so, that's something to fix on the Tectonic end.

**Comment by AlphaKretin** [comment]
image
That hypothesis seems correct. I also took the opportunity to test levels and SP. Team code `gGP/wAEpC0ACim////8AGw==`

**Comment by AlphaKretin** [comment]
> My test case for forms has been lycanroc, and the _form_ part seems to work (team code: `gGMaxqtwjKUpSm/xsFcAAg==`), but the ability is undefined. Possibly because it only has one? If so, that's something to fix on the Tectonic end.

Ah, right! The issue is that the ability index can be 1, even if it only has one legal ability. Of course in that case, we have to clamp it back to the first.

**Comment by AlphaKretin** [comment]
image
Oh you love to see it. Just moves to go.

**Comment by AlphaKretin** [comment]
After dinner I'll print the allMoves list for Espeon from both Tectonic and loadData and compare them

**Comment by AlphaKretin** [comment]
[AERIALACE, BODYSLAM, BRICKBREAK, CENTER, CHARM, CLEARSMOG, CONCENTRATE, CONFIDE, COVET, DEADEND, DEEPBREATHING, DIG,             DOUBLEEDGE, DOUBLEKICK, EARTHPOWER, ECHOEDVOICE, ENDURE, FACADE, FAKETEARS, FLAIL, FLOWSTATE, FOCUSENERGY, FOLLOWME, GIGAIMPACT, HEADBUTT, HEALBELL, HELPINGHAND, HOLDBACK, HOP, HYPERBEAM, HYPERVOICE, LASERFOCUS, LASTRESORT, LATHER, LAUNDER, LONGSHOT, MIASMA, MIRRORCOAT, MUDSHOT, MUDSLIDE, MUDSLING, NATURALGIFT, NULLPULSE, PAYDAY, PROTECT, QUICKATTACK, RECOVER, REFLECT, REST, RETALIATE, RETREAT, ROLEPLAY, ROUND, SCHEME, SHADOWBALL, SHARPEN, SLEEPTALK, SLUDGE, SMASH, SMOG, SNORE, SPITEFULCHANT, STAMPDOWN, STARDANCE, STELLARATE, STRAFE, SUBSTITUTE, SUPERPOWER, SWEETSELENE, SWIFT, TACKLE, TICKLE, TRUMPCARD, UNDERMINE, VANGUARD, WEATHERBURST, WISH, WORKUP, YAWN, BRAINFREEZE, CALMMIND, CENSOR, DISARMINGVOICE, DREAMABSORB, EXPLOIT, FLING, FUTURESIGHT, HEARTSWAP, HELIOPAUSE, LIGHTSCREEN, MAGICCOAT, MINDWARP, OCCULTATION, PSYCHESWITCH, PSYCHIC, PSYCHICFANGS, PSYCHUP, PULSAR, RISINGPOWER, SCRUTINIZE, ... 15 more items ] 
    [AERIALACE, BODYSLAM, BRICKBREAK, CENTER, CHARM, CLEARSMOG, CONCENTRATE, CONFIDE, COVET, DEADEND, DEEPBREATHING, DIG, DOUBLEDOWN, DOUBLEEDGE, DOUBLEKICK, EARTHPOWER, ECHOEDVOICE, ENDURE, FACADE, FAKETEARS, FLAIL, FLOWSTATE, FOCUSENERGY, FOLLOWME, GIGAIMPACT, HEADBUTT, HEALBELL, HELPINGHAND, HOLDBACK, HOP, HYPERBEAM, HYPERVOICE, LASERFOCUS, LASTRESORT, LATHER, LAUNDER, LONGSHOT, MIASMA, MIRRORCOAT, MUDSHOT, MUDSLIDE, MUDSLING, NATURALGIFT, NULLPULSE, PAYDAY, PROTECT, QUICKATTACK, RECOVER, REFLECT, REST, RETALIATE, RETREAT, ROLEPLAY, ROUND, SCHEME, SHADOWBALL, SHARPEN, SLEEPTALK, SLUDGE, SMASH, SMOG, SNORE, SPITEFULCHANT, STAMPDOWN, STARDANCE, STELLARATE, STRAFE, SUBSTITUTE, SUPERPOWER, SWEETSELENE, SWIFT, TACKLE, TICKLE, TRUMPCARD, UNDERMINE, VANGUARD, WEATHERBURST, WISH, WORKUP, YAWN, BRAINFREEZE, CALMMIND, CENSOR, DISARMINGVOICE, DREAMABSORB, EXPLOIT, FLING, FUTURESIGHT, HEARTSWAP, HELIOPAUSE, LIGHTSCREEN, MAGICCOAT, MINDWARP, OCCULTATION, PSYCHESWITCH, PSYCHIC, PSYCHICFANGS, PSYCHUP, PULSAR, RISINGPOWER, SCRUTINIZE, SEERSTRIKE, SKILLSWAP, TOTALCONTROL, TRICK, ZENHEADBUTT, PSYPIERCE, MINDWAVES, KICKSTART, BABYDOLLEYES, PSYBEAM, CONFUSERAY, PUMMEL, ECLIPSE, SCAM, BEATUP]
    
Comparing these lists (JS on top, Tectonic on bottom), it seems the only discrepancy (in range, at least) is the absence of DOUBLEDOWN in the JS list. Presumably it's a similar issue with other moves for other Pokemon. We may be missing some moves in the loading process.

**Comment by AlphaKretin** [comment]
DOUBLEDOWN is in fact entirely absent from moves.json despite being in moves_new.txt

**Comment by AlphaKretin** [comment]
```ts
                 if (line.startsWith("[")) {
                    const value = line.substring(1, line.length - 1);
                    if (value === "DOUBLEDOWN") {
                        console.log("hey what gives!");
                    }
                    pairs.push({ key: "Bracketvalue", value: value });
                }
```

A simple test, but it confirms that DOUBLEDOWN is never found by the parser. Is it skipping lines for some reason?

AlphaKretin:
...actually hold on this should be its own issue

AlphaKretin:
image
OK, I had a severe skill issue there (my local copy Pokemon Tectonic having outdated data), but after fixing that, there are still some discrepancies, though fewer.

AlphaKretin:
image
Yes! Ok!

AlphaKretin:
@Cincidial I think everything is working, so I guess it's time for the cleanup you were wanting to do.

src/app/data/teamExport.ts
```diff
@@ -136,8 +136,9 @@ function encodeChunk(version: VersionMap, view: DataView<ArrayBuffer>, byteOffse
         byteOffset++;
     }
     if (hasForm) {
+        const formId = data.species.forms[data.form].formId;
```
Cincidial, line 139: Just for reference, I did a quick check on this and the max int value using form id is still 27. These bits support up to the number 64. So all good still.

AlphaKretin: Cool, I sorta thought so. They're probably the same in many cases, though I am aware of exceptions since Pikachu's forms start at like 3 for some reason.

AlphaKretin:
Looking at switching the base64 encoding over to base64url as discussed, so I can test the parallel change on Tectonic's end, and running into issues *here* as well. `base64url` is listed as a valid buffer encoding in the TypeScript types, but when it comes time to actually *run* the code, I get `Error: Unknown encoding: base64url`. Augh.

AlphaKretin:
However, `Buffer.from(x,"base64")` may reportedly silently support Base64URL as well, in which case we only have to worry when encoding (if even then).

Cincidial:
Ah I'm working on this right now as well haha
I've got a function to convert between them going, just testing things out

AlphaKretin:
Ah, excellent, I'll leave you to it then
I've got a conversion going on the Tectonic end, but the test case doesn't actually seem to have any of the relevant characters lol
`gGMbheDYjKUpSiQgkDYAAMIhm4ylKUoiUdwwABNCivyMpSlKJKBoBgAZBhqTjKUpSjKwqF4tAACBgaKMpSlKJHCoYg0jA3BJjKUpSiISFI0A`

Cincidial:
Oh try using a blank team if possible. Or remove moves and stuff so that -1 values get put in.
I think the relevant characters are the ones at the end of the range, so just getting all 1's in the bits should do it

AlphaKretin:
Good call. Here's a single team member, missing a move `gGMRwzLKUApQCi___DwA`

Cincidial:
Huh, I think, maybe, another benefit of using the URL safe encoding is you can double click to highlight the whole thing without it breaking up. Like before if there were slashes it would become segmented. Neato

Cincidial:
image
Look correct?
Thanks for the code, I ugh, broke encoding trying to do something apparently lol, was debugging it and wasn't sure which it was. But if that's correct, well at least it's encoding thats ded

AlphaKretin:
Yep that's correct

Cincidial:
Ugh, I gotta go eat some food lmao.
I'll push some broken stuff that I have right now. Idk what happened to export but it's kinda dead
For example, this is console log of party mon `data.items`
image

Cincidial:
I;'ll take a look when back, but if you end up looking in the meantime cool (no worries though, focus on the tectonic side)

AlphaKretin:
> I;'ll take a look when back, but if you end up looking in the meantime cool (no worries though, focus on the tectonic side)

There's not a ton left to do on the Tectonic side until we have this all done and the URL loading implemented, so I'd be happy to take a look... wheeeen I don't have a massive headache

AlphaKretin:
I managed to export the usual charmer team just fine with the latest code, fwiw. Is your underlying data up to date?

AlphaKretin:
Yeah I've tested different data sources, null values, can't recreate any encoding error

Cincidial:
Just got back
I'm running with `npm run data` right now. I tried dev as well but same issues

Cincidial:
After export can you import that code?
That's what's failing for me

AlphaKretin:
> After export can you import that code? That's what's failing for me

Ah, that would've been smart. Indeed I cannot.

AlphaKretin:
Ah, actually, I *can* import the sample team (`gGMbheDYjKUpSgQgkDYAwiGbjKUpSgJR4DETQor8jKUpSgSgaAYZBhqTjKUpShKwqF4tAIGBooylKUokcKhiDSMDcEmMpSlKAhIUjQA`). But not the case with various nulls (`gGP_____UKUpSv____8A`).

Cincidial:
> Huh somehow the pokemon party data itself is just a move array? image

jk idk what was happening here, I restarted npm and it's not doing that now at least

Cincidial:
Ah this might be part of it
image

AlphaKretin:
> Ah this might be part of it image

Oops. Really easy to get those mixed up when copy pasting around. Annoying that Typescript doesn't flag a mismatch there.

Cincidial:
All g, that didn't solve it anyways

Cincidial:
Ohhh I know what it is, I saw before even trying it
I updated the move masking part to use a local function so that it handle underfined errors by using -1
I did _not_ remember to mask -1 after though lmao. So it's just filling all the bits with 1

Cincidial:
Yeah I think that did it. Should be good now. Ima just do a couple more things then I'll ping ya
One question though, I see `versions.json` is excluded from the gitignore, can I change it to be excluded like the other gen files?

AlphaKretin:
> Yeah I think that did it. Should be good now. Ima just do a couple more things then I'll ping ya One question though, I see `versions.json` is excluded from the gitignore, can I change it to be excluded like the other gen files?

It needs to be included in the repo because the data scripts will only download one version at a time, and we want both/all versions up to date so that any version of the website can correctly import any team code.

Cincidial:
Ohh yeah that makes sense.

src/app/teambuilder/page.tsx
```diff
@@ -164,6 +164,7 @@ const TeamBuilder: NextPage = () => {
     function importTeam() {
         try {
             setCards(decodeTeam(teamCode));
+            alert("Team imported successfully!");
```
Cincidial, line 167: I forgot I removed this, tried importing, spammed the button like "why is nothing happening" then realized that's exactly why this is here and re-added it lmao.

Cincidial:
@AlphaKretin I think I'm done the stuff I wanted to quickly do. Feel free to check things out on your end when you want
Getting it RealClean :tm: is something way down the road maybe

Cincidial:
Ah wait, I'm just gunna add something for importing an empty string

AlphaKretin:
> @AlphaKretin I think I'm done the stuff I wanted to quickly do. Feel free to check things out on your end when you want Getting it RealClean ™️ is something way down the road maybe

That addition you mentioned aside, sounds good!

Cincidial:
Ight, I think that's everything. Well at least that I can think of last minute

AlphaKretin:
Found an issue, sorry to say. ...Though I've immediately realised what the cause is.
With the team code `gGMRwzLKUApQCi___DwA`, it imports as Noibat with main data but Jellicent (correctly) with dev data. We cut Pokemon from the version map because we were using dex numbers, but uh... dex numbers can change between updates. And have. We don't need the version map for *exporting* (well technically we never need it for exporting except to save on redundant index calculations), but we do need it for importing.

AlphaKretin:
Empty import and forced ability selection are working though, as far as I can see

AlphaKretin:
OK, fixed that. Notably loading the same Pokemon on an older data version where it has different abilities, I can see the ability correctly switches to the appropriate ability for the index, as expected.

AlphaKretin:
With that, yeah, I think we're good! Eager to get this live.

PR #122, opened 2025-04-21
Autocomplete trainer movesets if none defined

### LINKED ISSUE #105: Autopopulate standard movesets for trainers with none defined
Signature move if applicable + last 3-4 level up moves

AlphaKretin:
Close #105 

WIP, results currently do not match up to in-game

### DIFF
--- a/src/app/data/types/Trainer.ts
+++ b/src/app/data/types/Trainer.ts
@@ -47,10 +47,12 @@ export class Trainer {
                 if (signatureMove) {
                     newMoves.push(moves[signatureMove]);
                 }
+                // don't auto-learn moves past level 50
+                const maxLevel = Math.min(mon.level, 50);
                 const movesUpToLevel = pokemon[mon.id].levelMoves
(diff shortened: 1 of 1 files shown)


AlphaKretin:
Checking Tectonic's code, it seems the move autopopulation caps at level 50 (perhaps to avoid including some of the awkward, situational moves learned at the end of learnsets). Let's try that and see if things are better behaved. Worst case scenario I'll try copying its logic more directly.

AlphaKretin:
image
Image
Now that's more like it! I'll just reverse the list so the order matches up to the game.

AlphaKretin:
This isn't finished yet though, Gastrodon's signature move seems to not be appearing
image
(also if you look closely abilities are wrong but that's a different issue)

AlphaKretin:
oh that's a silly mistake

AlphaKretin:
image
Ta-da! Think I'm happy with this now

PR #134, opened 2025-04-24
Load team codes from URL parameter

### LINKED ISSUE #91: Allow loading team code from URL parameter
https://developer.mozilla.org/en-US/docs/Web/API/Navigator/registerProtocolHandler

AlphaKretin:
Close #91 

Want to implement https://developer.mozilla.org/en-US/docs/Web/API/Navigator/registerProtocolHandler before considering complete.

### DIFF
--- a/src/app/teambuilder/page.tsx
+++ b/src/app/teambuilder/page.tsx
@@ -49,6 +49,25 @@ const TeamBuilder: NextPage = () => {
         }
     }, []);
 
+    useEffect(() => {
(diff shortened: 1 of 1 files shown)


AlphaKretin:
On second thought, `registerProtocolHandler` doesn't seem super appropriate here. Tectonic can already construct the appropriate URL itself, and choose between the main and dev branches of the website as necessary. Plus this seems like a headache to test when you have to be doing things on the right domain.

PR #137, opened 2025-04-24
Propagate trainer data across extensions

### LINKED ISSUE #77: Make trainer extension work properly
Cursed etc. trainers should fall back to the data from their uncursed versions if it doesn't exist.

AlphaKretin:
Close #77 

Inheriting flags seems to work, adding new Pokemon to a team doesn't seem to (I'm a bit baffled as to why), I haven't checked in detail whether or not new moves, items, style points and such on existing Pokemon are being applied properly.


AlphaKretin:
That was a silly mistake, but the excuse to step away for a second and look at it with fresh eyes helped catch the overriding issue, so I'll take it

Issue #63, opened 2025-04-16
Improve visual appearance of type matchup tables


AlphaKretin:
Visual style is in flux at the moment during this big Techo rework, but this particularly stands out as unfinished

Cincidial:
Is this for the big type chart, or the one in the modal?
If in the modal might be covered by that other PR

AlphaKretin:
It was the one in the modal I think, but it couldn't hurt to do a pass over the big one, as well as the tables in the team builder (in particular, the colour gradient for a "good"/"bad" number of matchups is really poorly executed).

Cincidial:
I'll pick this up as well unless you have interest. Also planning to add the hover text to the type chart cells.

I'll take a quick peak at the matchup table and see if ive got any ideas too

AlphaKretin:
Sounds good!

Cincidial:
Just thought I'd post the this is still in progress & a word of caution if you ever think of it:

Do not spend time trying to get complimentary colors for each pokemon type so the text displayed can be contrasting. Not only is it a huge hassle, but in the end literally just black or white looks significantly better lol. 

1.5 hours well spent? Maybe, maybe not. Lessons learned though for sure

AlphaKretin:
> Just thought I'd post the this is still in progress 

I'd been meaning to check in, thanks for mentioning it!

> 1.5 hours well spent? Maybe, maybe not. Lessons learned though for sure

Oh nooooooo

Issue #136, opened 2025-04-24
Optimise image cache usage on Vercel


AlphaKretin:
https://vercel.com/docs/image-optimization/managing-image-optimization-costs

Due to the large number of images, we incur high costs for image transformation. However, these should be one-time costs for each image, after that they're stored in the cache which has a much higher allowed usage.

*However*, depending on the website the traffic gets in the future, the number of images we have * the number of users could still exceed cache reads. It's possible to store images locally instead of in the cache, but this would of course incur data transfer costs instead. It's potentially an option to host them offsite and hotlink them, even just from the repo, but of course this circumvents Vercel's optimisations which should at least in theory be beneficial to both us and the user.

It's possible that upgrading to the Vercel Pro plan may just be idea - it would not only resolve this issue, but the issues we've been having with collaboration, allowing me to invite @Cincidial and any other future contributors to the project team properly. However, unless Tectonic has a marketing budget that can foot the bill, I'm not exactly keen to pay it out of pocket, I envisioned this as a cost-free hobby project (besides the domain which I already own for other reasons). 

Another option would be to host the website independently instead of using Vercel. This would require our own continuous deployment setup, but that's far from impossible. Again, this would likely depend on what resources Tectonic has available. I'm not sure if Tectonic has its own webserver or simply uses Wordpress or another similar host.

Cincidial:
I was reading those docs a bit more. 
I think as long as things get cached regionally they don't incur a "cache hit". So, maybe just be setting the cache max-age to over 1 month we should be good (assuming those regional caches respect it). Since the regional one is free to hit as well then many users shouldn't pose too much of an increase. 

In any case setting that cache age will mean users shouldn't cause many cache misses.
I don't think pro is worth it, as if who ever is paying leaves / no longer supports it then refinancing needs to be found...

As far as another site, for Techo I just used github io which seemed to do everything as desired. In general this seems like a fully static-content website with JS being done client side for stuff. Nothing really happens on the server end (unless there was a plan for a showdown esq battle sim).

AlphaKretin:
We'd love a battle sim to exist but that's definitely not in the scope of this project lol
github.io is definitely an option, I'm not planning any server-side activity, I simply went with Vercel because it's the simplest option for deploying a Next.js app, since they set it all up for you. Since this is a Typescript React app and not just html + js, idk if we can just use the standard Github Pages deployment process, though I do think there is a way, it just requires a little more work.

[...]

PR #138, opened 2025-04-26
#63: Improve Visuals of Type & Matchup charts. Anndd other things

Cincidial:
Yeah this definitely should of been split up, but I kinda just kept going with the flow while doing it. I'll try to summarize what's going on here:
- Fixed a bug with type calc matchup returning  0 instead of 2 for double dmg move? Just want to check this is right?
- Cleaned up how the type calcs method is called via adding classes and added a util for best move matchup
- Removed the comments used to stop tailwind from purging our css classes and replaced it with the inline source thing. It's pretty neat actually - via some specific string formatting it generates all the classes
- Collapsed the type badge / header classes into one and added an enum to generate the type of element you want for this. Currently supports a stacked dual type display, a row dual type display, a single capsule display and a col / row single type display
- Removed a bunch of extra classname values that were being overwritten / unneeded
- Added type chart tool tips
- Changed all matchup charts to display nothing if it were to display 0. This is to remove extra visual noise
- In general I tried to make the type and matchup tables smaller and require less scrolling without losing information
- Matchup tables in team builder show types to help with cases like the type changing item. I'd like to maybe add abilities as well (later in another PR)
- **I removed the bad -> good matchup highlighting** I like the idea, but wanted to see what it looked like without it. Basically letting the user form their own conclusion from the data instead of hinting. It's less noisy, but I like the idea of giving hints. Soo, take a look. Up to you, but no worries about adding it back if that's the direction. 
- Added a shadow to the type badge text to help out electric and ice! The other benefit as well, but not as much. In another PR I'd like to apply this to the main pokedex table to see if it helps highlight all the text

Note: I didn't do extensive testing yet, just checked around the pages and stuff. Wanted to finally put a stopping point to the changes in this and get some feedback


src/app/data/typeChart.ts
```diff
@@ -132,23 +146,28 @@ export function calcTypeMatchup(atk: AttackerData, def: DefenderData) {
...
+    if (atk.move != undefined && !isNull(atk.move) && atk.move.isAttackingMove()) {
+        const doubleDealtMatch = doubleDealtMoves.find((x) => x.move == atk.move!.id);
         if (
             doubleDealtMatch !== undefined &&
             (doubleDealtMatch.type1 == def.type1.id ||
                 doubleDealtMatch.type1 == def.type2?.id ||
                 doubleDealtMatch.type1 == thirdType)
         ) {
-            atkMoveCalc = 0;
+            atkMoveCalc = 2;
```
Cincidial, line 159: Just commenting to draw extra focus to this. I'm pretty sure it's a bug? Just want to make sure though

src/app/teambuilder/components/AtkTotalCell.tsx
```diff
@@ -20,44 +19,10 @@ export default function AtkTotalCell({
...
-                          },
-                          { type1: type }
-                      )
-                  )
-              )
-            : 1;
-        return compare(mult, total);
+        return compare(calcBestMoveMatchup(c, new DefenderData(type)), total);
     }).length;
-    // Updated color scheme with better readability and subtle transitions
```
Cincidial, line 40: See PR desc about this. In brief just wanted to show how it looks without the colors. I'm like the idea of count based coloring, but it's a tough one to indicate the meaning without being unclear.

Cincidial: Also, I'm probably a good test for this - I'm red + green color blind. So these colors would need to be tweaked anyways (:

src/app/teambuilder/components/MatchupMonCell.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+    return (
+        <td>
+            <div className="flex pr-2 space-x-1">
+                <Image
+                    src={c.species.getImage(c.form)}
+                    alt={c.species.getFormName(c.form) ?? c.species.name}
+                    title={c.species.getFormName(c.form) ?? c.species.name}
+                    height="80"
+                    width="80"
```
Cincidial, line 19: This a struggle between "image too small" and "table cells with numbers too tall"

src/app/teambuilder/page.tsx
Cincidial, line 1: The main idea with the matchup tables was
- Remove the need to scroll
- Make more compact
- Provide type clarity alongside the pokemon

Cincidial:
Might want to pull locally to view. Not sure if this vercel deployment is worth it (idk if this affects the counts!)
Also, here's the full teamcode from the team code stuff. It was a blessing to help test things

```
gGMbheDYjKUpSiQgkDYAAMIhm4ylKUoiUdwwABNCivyMpSlKJKBoBgAZBhqTjKUpSjKwqF4tAACBgaKMpSlKJHCoYg0jA3BJjKUpSiISFI0AAA
```

AlphaKretin:
Oh, don't worry, I always test locally.
I'll take a look over all of this later, but as a quick word on the "good"/"bad" colouration - I'm largely happy cutting it, it doesn't look great. It made more sense in the google sheet with conditional highlighting and all. In particular, having more resistances/super effective moves being "more good" isn't even objectively true, like if you have one immunity to a type on a team that might be all you need. Personally, when I use the spreadsheet, there are two particular things I'm looking for
- Types I have zero resistances to (especially if I have a weakness to the type on the team)
- Types I can't hit for super-effective damage
Maybe there's a way to flag those specifically without such an obtrusive and arbitrary colour scale feature.

Cincidial:
Also, one thing that might be cool to add an issue for:
On one of the reference sites I use for the matchup table stuff (looking at how they do it) the offensive coverage has a % of pokedex covered
image

Could be a neat way to provide some more team info
https://www.pkmn.help/offense/

AlphaKretin:
> Also, one thing that might be cool to add an issue for: On one of the reference sites I use for the matchup table stuff (looking at how they do it) the offensive coverage has a % of pokedex covered image
> 
> Could be a neat way to provide some more team info https://www.pkmn.help/offense/

Oh, yes, this would be cool! I have an issue for "double type matchup tables", with the understanding that it won't be super practical to do *literally* that - this would be a great substitute.

AlphaKretin:
Oh, by the way, it's not a huge deal, but you can check for build errors by running `npm run build` before committing/pushing

Cincidial:
oh yeah sorry, I keep forgetting to do that. You probably get spammed the failures from that too.
I'll fix the error from it

src/app/data/typeChart.ts
AlphaKretin: Yeah, it must be. No idea how that one happened.

src/app/teambuilder/components/AtkTotalCell.tsx
AlphaKretin: As described in the PR comments, I'm happy cutting this, and if I restore something like it, it'll be as a somewhat new solution that hopefully doesn't depend on distinguishing colours.

src/app/data/typeChart.ts
```diff
@@ -3,20 +3,34 @@ import { ExtraTypeMove } from "./moves/ExtraTypeMove";
...
 import { Ability } from "./types/Ability";
 import { Move } from "./types/Move";
+import { PartyPokemon } from "./types/PartyPokemon";
 import { PokemonType } from "./types/PokemonType";
+import { isNull } from "./util";
 
 export const typeChart = loadedChart;
 
-interface AttackerData {
+export class AttackerData {
```
AlphaKretin, line 12: I don't think this needs to be a class, unless it's going to have a utility method added to it. We're fine using interfaces elsewhere in the code, like MoveData.

src/app/globals.css
```diff
@@ -84,3 +84,5 @@ body {
     --color-barely-effective: #e3649f;
     --color-not-very-effective: #f7be81;
 }
+
+@source inline("{hover:bg,bg,to,from}-{fire,water,steel,rock,psychic,poison,normal,ice,ground,grass,ghost,flying,fighting,fairy,electric,dragon,dark,bug,mutant}{,-bg}");
```
AlphaKretin, line 88: It turns out, the reason I thought safelists didn't exist, *and* the reason I couldn't get this working before, was because I was on an outdated version of tailwind. They must have re-added the feature *really* recently, and when you wound up updating tailwind back when, that wound up bringing it in. Updating my reqs to the ones in the (since-modified) package.json makes it work.

src/app/pokedex/components/MoveDisplay.tsx
```diff
@@ -5,7 +5,7 @@ import { getSignatureMoves } from "@/app/data/signatures";
 import { Move, moveCategories, MoveCategory } from "@/app/data/types/Move";
 import { Pokemon } from "@/app/data/types/Pokemon";
 import { isNull } from "@/app/data/util";
-import TypeBadge from "@/components/TypeBadge";
+import TypeBadge, { TypeBadgeElementEnum } from "@/components/TypeBadge";
```
AlphaKretin, line 8: I've heard there are issues with Enums in typescript, so we might want to use another similar solution for this, but I should probably fact check that before actually acting on it.

AlphaKretin:
Would it be possible to make separate PRs for approved smaller, standalone changes, like the one to globals.css? That way I can merge them into main, merge main into this, and that will reduce the number of things this PR covers.

src/app/data/typeChart.ts
```diff
@@ -132,23 +146,30 @@ export function calcTypeMatchup(atk: AttackerData, def: DefenderData) {
             atkAbilityCalc = 1.3;
         }
     }
-    const atkMove = atk.move;
-    if (atkMove !== undefined) {
-        const doubleDealtMatch = doubleDealtMoves.find((x) => x.move == atkMove.id);
+
+    let atkMoveCalc = 1;
+    if (atk.move != undefined && !isNull(atk.move) && atk.move.isAttackingMove()) {
```
AlphaKretin, line 151: While I admit that semantically isNull seems like it should operate on objects exclusively, it does also already include an undefined check for situations like this. I need to remember that myself, more often.

src/app/data/typeChart.ts
```diff
@@ -132,23 +146,30 @@ export function calcTypeMatchup(atk: AttackerData, def: DefenderData) {
...
+        const doubleDealtMatch = doubleDealtMoves.find((x) => x.move == atk.move!.id);
```
AlphaKretin, line 152: It's probably not surprising that at some point in the future I want to replace these lists of special moves with move subclasses like ExtraTypeMove, by checking function codes, like I am with all other move functionality. But that doesn't have to be part of this PR, I'll get to them when they come up on the move spreadsheet.

src/app/pokedex/page.tsx
```diff
@@ -144,15 +143,15 @@ const Home: NextPage = () => {
...
-                    removeFilter={removeFilter}
-                    setCurrentFilter={setCurrentFilter}
-                />
-
                 <TabContent tab="Pokemon" activeTab={activeTab}>
+                    <FilterInput
+                        currentFilter={currentFilter}
+                        filters={filters}
+                        onAddFilter={handleAddFilter}
+                        removeFilter={removeFilter}
```
AlphaKretin, line 151: I had the FilterInput visible on all tabs because the other tabs could add to it - but I suppose since doing so tabs back over to the Pokemon tab anyway, this could be fine. I'll have to see how it feels in practice.

src/app/pokedex/page.tsx
```diff
@@ -279,40 +283,39 @@ const Home: NextPage = () => {
...
-                                    </th>
-                                    {realTypes.map((t) => (
-                                        <th
-                                            key={t.id}
-                                            className="px-4 py-2 text-sm font-medium text-gray-700 dark:text-gray-300"
-                                        >
-                                            <TypeBadge type1={t} />
-                                        </th>
+                    <table className="mx-auto border border-gray-300 dark:border-gray-700 bg-white dark:bg-gray-800">
+                        <thead className="">
```
AlphaKretin, line 287: If the class name is blank, we should be able to exclude it entirely. In any case, appreciate you cleaning this stuff up!

src/app/pokedex/components/MoveDisplay.tsx
AlphaKretin: Haven't done that research yet, but to explicitly state what an alternate solution might be: Something like the "player"/"opponent" or "weak"/"strong" switches elsewhere in the code - basic strings constrained by union types. Alternatively, if we decide Enums are fine, we might want to switch those examples over, for consistency.

src/app/teambuilder/page.tsx
```diff
@@ -350,136 +350,99 @@ const TeamBuilder: NextPage = () => {
...
+                            </tr>
+                        </thead>
+                        <tbody className="bg-white dark:bg-gray-800 divide-y divide-gray-200 dark:divide-gray-700">
+                            {validCards.map((c) => (
+                                <tr key={c.species.id}>
+                                    <MatchupMonCell key={c.species.id} c={c} useMoves={true} />
+                                    {realTypes.map((type) => (
+                                        <TypeChartCell
+                                            key={type.id}
+                                            atk={undefined}
```
AlphaKretin, line 424: I'm not sure I'm happy having to have all these `atk={undefined} def={undefined}` throughout the code just so there can be one case where they're defined. But I do see the logic of using the same Component for these two use cases. It should be possible to have a default value set so they can be excluded when not relevant, I think.

src/app/pokedex/components/MoveDisplay.tsx
AlphaKretin: OK, shockingly, it seems like the bad reputation of typescript enums were overblown clickbait. They're basically the same as the string union method under the hood, except that renaming the enum values is easier than with string constants. So I'm happy to keep this and potentially shift other examples over to enums in the future.

src/app/pokedex/page.tsx
AlphaKretin: Yeah I like this actually

src/components/TypeBadge.tsx
```diff
@@ -1,33 +1,61 @@
...
+    CAPSULE_STACK,
+    CAPSULE_ROW,
+}
+
 interface TypeBadgeProps {
-    type1: PokemonType;
-    type2?: PokemonType;
-    hyper?: boolean;
+    types: (PokemonType | undefined)[];
+    useShort: boolean;
```
AlphaKretin, line 14: The short names don't look great, aesthetically, but I understand the benefits of the horizontal compression. Maybe there are "nicer" looking abbreviations we can try? Or type symbols? But otherwise I'm fine with this.

AlphaKretin: Is there room for 5 characters? That would allow several types to be spelled out in full, and more readable abbreviations for many other types (e.g. NORM, FIGHT, ELEC)

src/app/teambuilder/components/MatchupMonCell.tsx
AlphaKretin: It may be aesthetically best here to use the menu icons here rather than the full sprites shrunken down, I just didn't want to upload *another* thousand images.

AlphaKretin:
The tables being reoriented is throwing me off, but that's entirely because I'm just used to the alternative, which was a better fit in Google Sheets. This definitely makes more sense functionally since the user will be scrolling down the page. 

Looking at the total cells, it *does* kind of just become a mess of numbers without the colours, but again I don't think the colours were a great way to go about it either. Something to work on in the future, I think.

AlphaKretin:
To go over each of the points in the PR.
> Fixed a bug with type calc matchup returning 0 instead of 2 for double dmg move? Just want to check this is right?

Yeah, this should be correct. No idea how I wound up putting a 0 in there. 
> Cleaned up how the type calcs method is called via adding classes and added a util for best move matchup

Not sure if we need classes here unless you anticipate adding helper methods in the future. 

> Removed the comments used to stop tailwind from purging our css classes and replaced it with the inline source thing. It's pretty neat actually - via some specific string formatting it generates all the classes

This ever not working was my own fumble, glad to have it.

> Collapsed the type badge / header classes into one and added an enum to generate the type of element you want for this. Currently supports a stacked dual type display, a row dual type display, a single capsule display and a col / row single type display

In favour of this, and I've dispelled my doubts about using enums. Might want to use enums elsewhere in the code for stylistic consistency and ease of refactoring, rather than using strings checked against type unions. I like the larger look of the badges.

> Removed a bunch of extra classname values that were being overwritten / unneeded
Glad to have my mess here cleaned up, there's been a lot of slapshop copy and pasting across different parts of the code. Hopefully this will make it easier in the future to identify shared stylings and turn them into classes or components.

> Added type chart tool tips

This is good.

> Changed all matchup charts to display nothing if it were to display 0. This is to remove extra visual noise

So is this.

> In general I tried to make the type and matchup tables smaller and require less scrolling without losing information

This is good, they're much smaller.

> Matchup tables in team builder show types to help with cases like the type changing item. I'd like to maybe add abilities as well (later in another PR)

I like this.

> I removed the bad -> good matchup highlighting I like the idea, but wanted to see what it looked like without it. Basically letting the user form their own conclusion from the data instead of hinting. It's less noisy, but I like the idea of giving hints. Soo, take a look. Up to you, but no worries about adding it back if that's the direction.

As discussed, happy removing the colours, but do think I want to find something else in the future to aid in at-a-glance analysis of the totals.

> Added a shadow to the type badge text to help out electric and ice! The other benefit as well, but not as much. In another PR I'd like to apply this to the main pokedex table to see if it helps highlight all the text
I didn't even notice this going over it, so it's definitely not obtrusive. In favour. If anything it maybe doesn't quite help enough, the white text on light colours still isn't the most pleasant on the eyes. 

tl;dr, the broad strokes are all good except maybe the use of classes in damage calculation. Only things to clear up are the individual nitpicks in line comments.

src/app/data/typeChart.ts
Cincidial: Agreed, will revert this. 
As I converted more of them it started to feel worse and worse lol

src/app/globals.css
Cincidial: Nice, versioning do be funny like that. 

I decided to try it again because I was adding a ton more color classes doing stuff with the table, most of that didn't make it in the end, but I kept this thing. The old comment list had gotten even more ridiculous.

src/app/pokedex/page.tsx
Cincidial: Oh my bad, I had forgotten this in the summary list.
Will just add that the reasoning behind the change was to introduce more vertical space, since as you said, adding a filter goes to the pokemon tab anyways.

src/app/pokedex/page.tsx
Cincidial: Ah yeh my bad there, I think I was testing out making that top left corner have no background and forgot haha. 
Thanks for the catch!

src/app/teambuilder/page.tsx
Cincidial: Done! Aparently to get default props working the interface props cannot be type defined like `atk: PokemonType | undefined` it must be `atk? : PokemonType`. Then the inline defaults will work

src/components/TypeBadge.tsx
Cincidial: Honestly symbols would be really nice if they exist in tectonic. 
Something like this (just stolen from google images). Though might be less immediately obvious what the type is
image

I'll try some stuff out and see how it looks

src/app/teambuilder/page.tsx
AlphaKretin: Sweet

src/components/TypeBadge.tsx
AlphaKretin: Yeah, the fact they're not in Tectonic *could* hurt comprehension a little, but I think these icons (or some very similar to this) come from canon since gen 8, so there's sufficient precedent.

Cincidial: Ah, I thought I had remembered seeing them somewhere in a game. That would make sense.
I'll leave icon exploration for another PR and just quick check out text changes for now

Cincidial: Actually jk I'm going to try the icons out and see how it feels lol

Cincidial: Hmm jkjk again, I'm too worried about introducing something with copy right or something like that. It's probably a discussion with the devs if y'all want icons / how to source them

AlphaKretin: I meaaaaan we don't exactly have explicit permission to use any of our other images lol, fan project moment, but I see your point, couldn't hurt to wait and double check since it's not an asset otherwise in Tectonic

Cincidial: 5 Chars with upper
image

5 as regular
image

3 as regular
image

I think the main issue is the upper case doesn't look good, but for 3 to work lower case looks very odd.

Cincidial:
Thanks for reviewing all that! I know it was a lot haha. 

I also like the idea of using the icons instead of the full images to help with the matchup tables in the team builder. Was also worried about adding all the images though.

src/components/TypeBadge.tsx
AlphaKretin: Mm, yeah, it's tough

AlphaKretin:
> Thanks for reviewing all that! I know it was a lot haha.
> 
> I also like the idea of using the icons instead of the full images to help with the matchup tables in the team builder. Was also worried about adding all the images though.

We can probably just go ahead and if it causes problems that's just more incentive to think about #136. If they're displayed at native size they might not need any of what Vercel considers image transformations

src/components/TypeBadge.tsx
Cincidial: `Flyin` is really funny to me for some reason

AlphaKretin:
> I'll actually leave that for another PR if that's okay?

Yep for sure. This one's big enough without 1000 extra images :P

src/components/TypeBadge.tsx
Cincidial: Ah idk, I'm honestly down to just change it to whatever for now then get icons in there and see how it looks. I feel like those will be the best of all worlds - small width & descriptive. 

I'd also like to attach the icons to the capsule badges too. Since the text would be attached it could act as bit of a bridge to learn the icons via other places on the site

AlphaKretin: Oh yeah, putting them in the capsules would be good. I think that's even how they look in the games that have the icons.

AlphaKretin: We might want to colour pick the theme colours for the types to match the icons, too

src/app/teambuilder/components/MatchupMonCell.tsx
AlphaKretin: If we're adding icons in a future PR, we can consider this resolved for now

src/components/TypeBadge.tsx
AlphaKretin: Since this is also going in another issue/PR, I'll resolve this discussion too

AlphaKretin:
OK, are we good? I think we might be good

Cincidial:
I think it's good, nothing on my end to go in here. Everything else is other tickets

PR #143, opened 2025-04-26
Refactor Move subclass instantiation

### LINKED ISSUE #142: Refactor move instantiation
Not loving the way things currently work

Since the constructor already has access to the move's function code, it can index into any necessary maps to get the corresponding variable, rather than needing to pass that to the constructor.

That way, the argument to a Move constructor is always just a LoadedMove, which may allow for things to be simplified/genericised in some way.

Additionally, maybe the moveCodes array could be a static property on the class (with the map being separate if necessary).

Then we could loop over a list of all Move subclasses and have one block with `if functioncode in Class.movecode, new Class(move)`, instead of writing it out separately for every subclass

AlphaKretin:
Close #142

### DIFF
--- a/src/app/data/moves.ts
+++ b/src/app/data/moves.ts
@@ -18,31 +18,30 @@ import { Move } from "./types/Move";
 
 function loadMove(move: LoadedMove): Move {
     if (move.functionCode in multiHitMoveCodes) {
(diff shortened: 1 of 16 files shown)


AlphaKretin:
At a quick test, special move functions do still in fact work (specifically what I checked was that echoed voice still had the "turns" input)

PR #151, opened 2025-04-26
Implement Heavy Slam

AlphaKretin:
Currently bugged. Power seems to be *higher* for heavier targets, and is not a multiple of 5 despite the last step being a multiplication by 5 after a floor. No idea what's going on, but I'm on an early morning roll through easy tasks so I'm leaving this to investigate later.


AlphaKretin:
OK I went to test it again and it's behaving normally. I'm gonna assume this was some "didn't have files properly saved" issue or something

Issue #128, opened 2025-04-22
Check consistency on handling of null types


AlphaKretin:
I have a feeling there's some inconsistent between an undefined type using the `nullType` constant vs just `undefined` (as in `Pokemon.type2`). It would be good to standardise this.

AlphaKretin:
Reminder to self that the `isNull` helper method *does* include an undefined check. We should probably be using it anywhere the thing we're checking is maybe an object, and then in the future if we do want to change between `undefined` and `null[Object]` as our signifier of an non-existent entity, that would be a more behind the scenes change that doesn't require updating any logic.

AlphaKretin:
Ah, small problem with this idea - `!isNull` doesn't work as a type guard the way `!== undefined` does. It might be possible to specifically tell typescript the function is a typeguard, but I'm having trouble figuring out exactly how

AlphaKretin:
Wait, I'm getting somewhere...

PR #163, opened 2025-04-27
Make object null checking consistent

AlphaKretin:
Related to #128.

Always use `isNull` rather than comparing to undefined. This will allow more flexibility in the future depending on how we decide to handle null object values. This PR does not include actually making a decision on that, where things are currently somewhat inconsistent between representing an unselected object as undefined or the corresponding null object.

Leaving PR open for a bit to double check that I didn't miss any instances. I searched the codebase for `= undefined` (which should include both `== undefined` and `!= undefined`), but obviously there are false positives where variables are primitives and an undefined check *is* appropriate, so it would be easy for one to have blended in. There may also be another syntax I'm not thinking of that would escape this search.

### DIFF
--- a/src/app/data/typeChart.ts
+++ b/src/app/data/typeChart.ts
@@ -92,7 +92,7 @@ export function calcTypeMatchup(atk: AttackerData, def: DefenderData) {
 
     let defType2Calc = 1.0;
     let defAbilityCalc = 1.0;
-    if (def.type2 !== undefined) {
+    if (!isNull(def.type2)) {
         const defType2 = def.type2;
         defType2Calc = typeChart[atkType.index][defType2.index];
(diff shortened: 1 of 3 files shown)


AlphaKretin:
I definitely feel like I'm missing some here, like there should be more instances of checking `type2` against undefined...
Oh, I bet it's just like `if (pokemon.type2)` huh. that'll be annoying to search for

AlphaKretin:
Sooner or later I need to actually do something with this or just give up on it

AlphaKretin:
Closing this (for now?), #180 changes how null entities work and in general I'm not making progress on this PR as I keep making other changes that leave it outdated. May revisit in the future

Issue #147, opened 2025-04-26
Add type icons


Cincidial:
Add type icons to be used 

1. As the col header for tables
2. Attached to capsule badges with their text.

Also:

              We might want to colour pick the theme colours for the types to match the icons, too

_Originally posted by @AlphaKretin in https://github.com/AlphaKretin/tectonic-tools/pull/138#discussion_r2061400898_

Cincidial:
I'll take up this one next.

Cincidial:
Closing this as I think it was done in https://github.com/AlphaKretin/tectonic-tools/pull/168

PR #168, opened 2025-04-28
#147: Add Type Icons

Cincidial:
Added type icons from a MIT repo. 
To save on pictures I embedded them in the code as react components. So nothing actually gets served via public. They are SVGs so there shouldn't need to be optimization anyways.

I added the type icon to the badges, for type matchup tables and on the pokemon card to save space. I think I got all the spots, but let me know if I missed any!

I also updated the base colors to match this and some background ones for contrast. I'm by now means very good at the color stuff, so feel free to modify as needed. 

Also let me know if you see any odd looking capsules. For some reason the single capsule variant had to be converted to flex in order to display properly.


src/app/pokedex/components/PokemonModal.tsx
```diff
@@ -304,10 +304,7 @@ const PokemonModal: React.FC<PokemonModalProps> = ({ allMons, pokemon: mon, hand
                         </TabContent>
                         <TabContent tab="Def. Matchups" activeTab={activeTab}>
                             <div>
-                                <div>
-                                    <h3 className="inline font-semibold text-gray-800 dark:text-gray-100">
-                                        Defensive Matchups{defMatchupDifferentForAbilities ? " with:" : ""}
-                                    </h3>
```
Cincidial, line 310: You're already clicked on this tab, so no title needed

src/app/teambuilder/components/MatchupMonCell.tsx
```diff
@@ -1,42 +1,33 @@
...
                 <Image
                     src={c.species.getImage(c.form)}
                     alt={c.species.getFormName(c.form) ?? c.species.name}
                     title={c.species.getFormName(c.form) ?? c.species.name}
-                    height="80"
-                    width="80"
-                    className="m-1"
+                    className="w-20 h-20"
+                    height="192"
+                    width="192"
```
Cincidial, line 16: Note: height and width are actually meant to define the resolution of the image for aspect ratio calculation, not the size on the page

src/components/type_icons/LICENSE
Cincidial, line 1: This should be all that's needed for usage of the SVGs - just having this license here

AlphaKretin:
Somehow I missed this yesterday, will take a look, ty!

AlphaKretin:
Looks good overall, like the use of the icons in the move list as a way to save space. I'm glad the icons can fit into the rounded end of the badge, that looks great. And the original reason to add them, the matchup table headers, look fantastic. Very happy. Only little quibble is that the icons and the type names look somewhat vertically offset - idk if that's the drop shadow's fault? But the text looks higher in the badge compared to the icon.

src/app/teambuilder/components/MatchupMonCell.tsx
AlphaKretin: Ooooh. There might be other places we need to change that too, then (but not necessarily in scope here)

Cincidial: I didn't realize it either haha, was looking at the docs and noticed it

Cincidial:
> Looks good overall, like the use of the icons in the move list as a way to save space. I'm glad the icons can fit into the rounded end of the badge, that looks great. And the original reason to add them, the matchup table headers, look fantastic. Very happy. Only little quibble is that the icons and the type names look somewhat vertically offset - idk if that's the drop shadow's fault? But the text looks higher in the badge compared to the icon. I'd love if we could smooth that out, but no biggie if it causes issues.

For the vertical offset, yeah it's caused by the background of the icon. Since we are making the capsule the same color you can't see the background anymore, but if you change the colors the text becomes vertically aligned. But, now that I think about it, if the icon is the same height as the full capsule anyways, then the text should still be centered, as the icon within it's background is centered. 

I'll play around with it a bit

AlphaKretin:
I rushed to merge so I could go to bed and now I've missed a commit and the dev build is broken because it... can't fetch the data??? awaaaaa

AlphaKretin:
ok we're good i think

PR #180, opened 2025-05-03
Data Load Cleanup & Other Cleanup

Cincidial:
Yeah this kinda is very large. My running goals while doing these were:
- Remove files had the same name, for example `pokemon.ts`, `pokemon.ts` and `Pokemon.ts`. Also limit namespace pollution with same name stuff
- Separate data loading out into it's own area as it's for a script run pre-build
- Move inflated/hydrated data into it's own folder to separate it from the other stuff in data
- Get the evolution tree into the pre-load json
- By extension with the evo tree, this is also enabled signatures to be there and use of NTree to remove the duplicate code in the preload for handling trees
- Reduce number of files needed to be loaded from public to save on vercel hits

Of course this is a lot of change from what you had originally - no requirement to actually merge this just cause it's a big bundle of stuff. Some of it being a different style than before (for example more classes for data loading as I found it helpful). Lemme know your thoughts

I checked the pages for smoke test + a bit more. Haven't super in depth tested though. 
With the team build I think it's okay, but also I was using an old code and not getting what I expected. Not sure if that's cause I just don't remember what the code was for, if versioning is messed up, or something else. With a new code though it seems fine. 

Also see https://github.com/AlphaKretin/tectonic-tools/issues/135


AlphaKretin:
I haven't looked over this yet but immediately I want to say that I'm definitely in support of the goals laid out in the PR comment. I'm definitely eager to merge something like this if all the details look good.

AlphaKretin:
Currently marking off files that are only import/structure changes to narrow down what I have to look at, commenting on what I've noticed in the process - I like the changes to have TectonicData be a central thing, and I like the null entities being static properties of the class.

AlphaKretin:
I like the use of default field values to define the null entities also.

src/app/data/tectonic/PokemonType.ts
```diff
@@ -0,0 +1,29 @@
...
+        this.id = loaded.key;
+        this.index = loaded.index;
+        this.name = loaded.name;
+        this.weaknesses = loaded.weaknesses;
+        this.resistances = loaded.resistances;
+        this.immunities = loaded.immunities;
+        this.isRealType = loaded.isRealType;
+    }
+
+    getShortName(): string {
```
AlphaKretin, line 26: Do we still use shortnames now that we have icons? Not relevant to this PR to delete it, just noticed while looking through.

AlphaKretin:
image
I need a nap (wow the sky must be blue) but now I can focus on all of the more significant changes later.

src/app/data/tectonic/PokemonType.ts
Cincidial: Nope, that's a good point. Probably won't need it all as for short stuff the icon is preferable.

Cincidial:
> image I need a nap (wow the sky must be blue) but now I can focus on all of the more significant changes later.

Thanks for taking a look and take your time - there's quite a bit. Enjoy the weekend while it's here!

src/app/data/abilities/TwoItemAbility.ts
```diff
@@ -26,17 +23,11 @@ export const twoItemAbilities: Record<string, validateItemFunction> = {
...
-        this.validate = validate;
-    }
-
     public validateItems(items: Item[]): boolean {
         // allow anything if an item remains null, since constraints should only apply if both items are (to be) selected
         if (items.some((i) => isNull(i))) {
             return true;
         }
-        return this.validate(items);
+        return twoItemAbilities[this.id](items);
```
AlphaKretin, line 31: Was there a specific structural or optimisation reason for this change? We use similar patterns in some items and moves I believe. Something about having the validation function be a property of the instance "feels" right to me, but if there's a good reason not to I'm open to it.

src/app/data/tectonic/Pokemon.ts
```diff
@@ -58,67 +54,93 @@ function getterFactory<T extends keyof Pokemon>(mon: Pokemon, key: T) {
...
-        this.tribes = mon.tribes.map((t) => tribes[t]);
-        this.height = mon.height;
-        this.weight = mon.weight;
-        this.kind = mon.kind;
-        this.pokedex = mon.pokedex;
-        this.items = mon.wildItems.map((i) => items[i]);
-        this.evolutionTree = mon.evolutionTree!;
+        this.abilities = loaded.abilities.map((a) => TectonicData.abilities[a]);
+        this.levelMoves = loaded.levelMoves.map(([level, id]) => [level, TectonicData.moves[id]]);
+        this.lineMoves = loaded.lineMoves.map((m) => TectonicData.moves[m]);
```
AlphaKretin, line 103: This is out of the scope of the PR but since we use this pattern a lot I find myself idly wondering if Javascript has like... operator overrides, so we can define an array index method that automatically performs this mapping. Probably unnecessary just to save an extremely idiomatic oneliner lol.

src/app/data/tectonic/TectonicData.ts
```diff
@@ -0,0 +1,172 @@
...
+import { EncounterMap } from "./Encounter";
+import { Item } from "./Item";
+import { Move } from "./Move";
+import { Pokemon } from "./Pokemon";
+import { PokemonType } from "./PokemonType";
+import { Trainer } from "./Trainer";
+import { TrainerType } from "./TrainerType";
+import { Tribe } from "./Tribe";
+
+const data = loadedData as unknown as LoadedDataJson;
```
AlphaKretin, line 56: This workaround looks heinous. I assume it's necessary because of some underlying jank that isn't our fault, but could you explain it?

src/app/data/tectonic/TectonicData.ts
```diff
@@ -0,0 +1,172 @@
...
+    moves: undefined!,
+    items: undefined!,
+    heldItems: [],
+    pokemon: undefined!,
+    forms: undefined!,
+    trainers: undefined!,
+};
+
+TectonicData.moves = fromLoadedMapped(data.moves, (x) => {
+    const subclass = moveSubclasses.find((sc) => sc.moveCodes.includes(x.functionCode));
```
AlphaKretin, line 153: I like this pattern!

src/app/data/tectonic/TectonicData.ts
```diff
@@ -0,0 +1,172 @@
...
+    tribes: fromLoaded(data.tribes, Tribe),
+    abilities: fromLoadedMapped(data.abilities, (x) =>
+        x.key in twoItemAbilities ? new TwoItemAbility(x) : new Ability(x)
+    ),
+    trainerTypes: fromLoaded(data.trainerTypes, TrainerType),
+    encounters: Object.fromEntries(
+        Object.entries(data.encounters).map(([k, v]) => [k, { ...v, id: v.key.toString() } as EncounterMap])
+    ) as Record<string, EncounterMap>,
+    typeChart: data.typeChart,
+    moves: undefined!,
```
AlphaKretin, line 144: Similar to above, ideally type forcing like this isn't necessary (though obviously it sometimes is), when I get around to poking through the code myself I'll experiment with what happens if we lack the !

src/app/data/tectonic/TectonicData.ts
```diff
@@ -0,0 +1,172 @@
...
+    return subclass ? new subclass(x) : new Item(x);
+});
+Item.NULL = new Item();
+
+TectonicData.pokemon = fromLoaded(data.pokemon, Pokemon);
+Pokemon.NULL = new Pokemon();
+TectonicData.forms = fromLoadedArray(data.forms, Pokemon.loadForm);
+
+TectonicData.trainers = fromLoaded(data.trainers, Trainer);
+Trainer.NULL = new Trainer();
```
AlphaKretin, line 169: If NULL is going to be a static property of the class, would it not be better to define it on the class? Or does it not like calling the constructor like that in the class definition, I could easily imagine that.

src/app/data/types/NTreeNode.ts
```diff
@@ -1,7 +1,29 @@
+export class NTreeArrayNode<T> {
```
AlphaKretin, line 1: Just saying this out loud to reinforcing my understanding, this is a helper for converting an array of items into a tree? What's the purpose of it having a constructor and properties that aren't used in any methods? I guess I'll see where else it's used as I scroll through.

src/preload/loadTectonicRepoData.ts
```diff
@@ -0,0 +1,286 @@
...
+    // Fetch the data and load it into a usable format
+    const version: string = await handleFiles(
+        ["Plugins/_Settings/GameSettings.rb"],
+        (f: string[]) => parseVersionFile(f[0]),
+        dev
+    );
+    const loadedData: LoadedDataJson = {
+        version: version,
+        types: await handleFiles(["PBS/types.txt"], (f: string[]) => parseStandardFile(version, LoadedType, f), dev),
+        tribes: await handleFiles(
```
AlphaKretin, line 195: I wonder if having `await` on Object properties like this would be parallelised in the same way as Promise.all... likely not, but also likely not a huge deal.

src/preload/tectonicFileParsers.ts
```diff
@@ -0,0 +1,129 @@
...
+
+    if (dev) {
+        version += "-dev";
+    }
+    return version;
+}
+
+export function parseStandardFile<T extends LoadedData<T>>(
+    tectonicVersion: string,
+    ctor: new () => T,
```
AlphaKretin, line 29: the nomenclature "ctor" is throwing me. What does it mean?

AlphaKretin:
Alright, I've looked over all the code and made the comments I have, next step is to poke around with the things I said I would, investigate the potential inconsistency with team codes, and just generally test

src/app/data/tectonic/TectonicData.ts
AlphaKretin: Testing it myself, it looks like the issue is without the `unknown`, every entity has a `populateMap` field which in the preload is filled with functions. While it's not like it tries to save the functions to the JSON, this is still a waste of space. We should modify the JSON writing to exclude the `populateMap` field.

AlphaKretin: Oh, actually while the `populateMap` field being there *is* an issue worth fixing, the actual syntax error is that the `populate` method of course doesn't get saved at all to the JSON. So yeah, the classes should be crunched down to interfaces for saving and later loading.

src/app/data/tectonic/TectonicData.ts
AlphaKretin: I see, it's to override the type signature not matching. Is there a reason these are left until later besides code tidiness?

src/app/data/types/NTreeNode.ts
AlphaKretin: I see it is used elsewhere actually, I need to spend a sec getting my head around it

src/preload/loadTectonicRepoData.ts
AlphaKretin: It is in fact not a huge deal but it *does* feel like the preload takes a bit longer to run

AlphaKretin:
> With the team build I think it's okay, but also I was using an old code and not getting what I expected. Not sure if that's cause I just don't remember what the code was for, if versioning is messed up, or something else. With a new code though it seems fine.

I tested using the example team and was able to load 4 different codes without a problem. However, the fact they *are* 4 different codes, instead of two, is maybe a problem, as the code generation seems to have changed in this branch.

main, 3.2.4
```
BEMYgdDEjKUpSgQhDCcbgfEtjKUpSgJRnG0RQsp9jKUpSgSggA4TBKoljKUpShKwsEgtAMFxMoylKUokcJxeDSMEM5yMpSlKAhIMjQA
```
main, 3.3.1-dev
```
hGMbxfDYjKUpSgQglDcAwiGdjKUpSgJR5DETQor-jKUpSgSgaAYZBhqVjKUpShKwqF4tAIGBpIylKUokcKhiDSMDcEmMpSlKAhIUjQA
```
pr, 3.2.4
```
BEMYgdDEjKUpSiQhDCcAG4HxLYylKUoiUZxtABFCyn2MpSlKJKCADgATBKoljKUpSjKwsEgtAADBcTKMpSlKJHCcXg0jBDOcjKUpSiISDI0AAA
```
pr, 3.3.1-dev
```
hGMbxfDYjKUpSiQglDcAAMIhnYylKUoiUeQxABNCiv6MpSlKJKBoBgAZBhqVjKUpSjKwqF4tAACBgaSMpSlKJHCoYg0jA3BJjKUpSiISFI0AAA
```

src/app/data/abilities/TwoItemAbility.ts
Cincidial: Honestly, was no reason for this. When I was doing the import fixes from file location changes I was kinda just on auto pilot changing little things. For this one I saw the property wasn't necessary and was a slight deviation from the other patterns so changed it. 

No worries if the other way feels more right. Can revert!

src/app/data/tectonic/Pokemon.ts
Cincidial: Oh yeah that would be nice. I'm not sure myself. I know we could define functions on everything for that, not sure if it would be worthwhile at that point though?

src/app/data/tectonic/TectonicData.ts
Cincidial: Yeah you found it.
The json stringify saves some stuff that cannot be recreated. So casting to unkown first is the way around that. 
I wasn't sure if there was a way to change the stringify portion via overrides, if so that would help here.

src/app/data/tectonic/TectonicData.ts
Cincidial: Oh yeah this serves 2 purposes:
1. As you mentioned get around the signature not matching
2. Everything with undefined requires some other portion of `TectonicData` to already be populated before it itself can be. For example, Moves and Pokemon both use the `TectonicData.types` info`. However, you cannot access the global before it's fully constructed, thus these cannot be inline otherwise we can't compile. They also need to be done in a specific order, as for example Pokemon uses moves

src/app/data/tectonic/TectonicData.ts
Cincidial: Since the NULL field is `static` JS performs both the declaration and assignment immediately on the first time the class is used. Which I thought was causing issues with other import dependencies not being loaded, but honestly after fixing the other compiling issues I never reverted this one to test if it was actually needed. I'll try it out, as I too much prefer the assignment to be inline with the declaration.

src/preload/loadTectonicRepoData.ts
Cincidial: No I don't think it's parallelized at all. My intent with this approach was to avoid the file indexing requirement from `Promise.all`. However, at the time I also didn't have the overall type defined (as I didn't need it at that point). With it defined now I can probably bring back the promise all and avoid the indexing at the same time

src/preload/tectonicFileParsers.ts
Cincidial: `ctor` & not used here, but `dtor` are old (though I think still used? Idk `back in the day™` they were haha) c++ slang for constructor and destructor. This parameter is asking for an argument that is the constructor function of the type so it can also call `new` on it.

Cincidial:
> > With the team build I think it's okay, but also I was using an old code and not getting what I expected. Not sure if that's cause I just don't remember what the code was for, if versioning is messed up, or something else. With a new code though it seems fine.
> 
> I tested using the example team and was able to load 4 different codes without a problem. However, the fact they _are_ 4 different codes, instead of two, is maybe a problem, as the code generation seems to have changed in this branch.
> 
> main, 3.2.4
> 
> ```
> BEMYgdDEjKUpSgQhDCcbgfEtjKUpSgJRnG0RQsp9jKUpSgSggA4TBKoljKUpShKwsEgtAMFxMoylKUokcJxeDSMEM5yMpSlKAhIMjQA
> ```
> 
> main, 3.3.1-dev
> 
> ```
> hGMbxfDYjKUpSgQglDcAwiGdjKUpSgJR5DETQor-jKUpSgSgaAYZBhqVjKUpShKwqF4tAIGBpIylKUokcKhiDSMDcEmMpSlKAhIUjQA
> ```
> 
> pr, 3.2.4
> 
> ```
> BEMYgdDEjKUpSiQhDCcAG4HxLYylKUoiUZxtABFCyn2MpSlKJKCADgATBKoljKUpSjKwsEgtAADBcTKMpSlKJHCcXg0jBDOcjKUpSiISDI0AAA
> ```
> 
> pr, 3.3.1-dev
> 
> ```
> hGMbxfDYjKUpSiQglDcAAMIhnYylKUoiUeQxABNCiv6MpSlKJKBoBgAZBhqVjKUpSjKwqF4tAACBgaSMpSlKJHCoYg0jA3BJjKUpSiISFI0AAA
> ```

Hmm yeah I thought something felt off. I think the version indexing got mapped slightly differently somehow and that's causing the index numbers to be different from the main branch but same version export. But, since the export & reimport is working there shouldn't be any missing info. Will look into it more

src/app/data/tectonic/TectonicData.ts
Cincidial: Ah unfortunately still no - since the default property values may include a reference to TectonicData. For example with moves `type: PokemonType = TectonicData.types['Normal']` we cannot include the assignment with the declaration.

src/app/data/types/NTreeNode.ts
Cincidial: Though some of the functionality is in the tree part (the going to the array) **

src/preload/loadTectonicRepoData.ts
Cincidial: Nice, looks like this can be done without indexing, while keeping the `Promise.all`

Cincidial:
Okay, yeah if I diff a prettifyied json of the versions then I get some differences between the movesets. Seems like it's all for pokemon like caterpie / wurmple / etc and their lines though...  But even though they are different codes it still loads everything just fine. So I'm not sure if a bug may have been fixed with their learnsets not being propagated in versions?

Aha, there it is I think. Something to do with the new NULL statics and comparing them
image
Item1 type is being set to true when it shouldn't

Cincidial:
What caused the difference in the first place with the type `id` I actually have no idea. Reading the code it looks like the change is just fine, but somehow the old code `nullType` was using `NORMAL`? That's my only guess, I have no idea where that would have been set though as the declaration of `nullType` uses just an empty string like the new one.

src/app/data/tectonic/Pokemon.ts
AlphaKretin: Probably not lol. Anyway, just an idle thought.

src/app/data/tectonic/TectonicData.ts
AlphaKretin: Ohhh, right, yeah that's a pretty good reason to leave them until later. A pattern we've used in similar cases (like code loading I think?) is to start with a `Partial<T>` and then after filling in all the blanks reassign it to a `T` and return/export it. I'll try implementing that in this case and see if anything breaks.

src/app/data/tectonic/TectonicData.ts
AlphaKretin: Ah, right, that makes sense.

src/app/data/abilities/TwoItemAbility.ts
AlphaKretin: Yes pls

AlphaKretin:
The reason itemTypes were defaulting to normal is because that is the behaviour that matches Tectonic - an item type is stored even if a Pokemon isn't holding a type changing item.

src/app/data/tectonic/TectonicData.ts
AlphaKretin: OK I can't find a remaining example of that pattern in the current code, and it doesn't like it because TypeScript can't tell we *have* defined these elements on this particular instance and just looks at the possible undefineds in the type signature. Hrmgh.

AlphaKretin: But we can assert that they exist in this context, which feels better than forcing undefined values onto a type signature that doesn't support them.

src/app/data/tectonic/TectonicData.ts
AlphaKretin: OK so the overall goal here is:
- Don't try to save populateMap or populate to JSON
- Don't expect them when loading
I'm thinking having separate types for LoadedData in the preload context and the running context. The latter can just be `Omit<former,"populate">` pretty much.

AlphaKretin:
The build error is... with the transpiled javascript???

AlphaKretin:
> The build error is... with the transpiled javascript???

OK if I run the website locally I get `Error: can't access lexical declaration 'TectonicData' before initialization` which... ugh. What? My change shouldn't have affected the order these files are loaded. Or made anything circular- OH NO OF COURSE it's the null move constructors and stuff. That does need Tectonic Data to already be... right, ok. I see. Yeah I don't think there's a way around this.

src/app/data/tectonic/TectonicData.ts
AlphaKretin: OK, nope, nevermind, that was silly of me. As you say, we need TectonicData to already exist and be exported for the definition of some later parts to work properly. I don't love this situation, but the alternative would be to have all of the data records separate again, and I see the structural advantage of bundling them like this. I'll relent, this is the way to go.

src/app/data/types/NTreeNode.ts
AlphaKretin: > Yup it's a helper that goes both ways - converts a tree to an array and vice versa. This is used to get around the cyclical properties introduced by tree nodes having both parent and child when exporting to JSON.

Ah, this makes sense. OK I'm satisfied with that explanation 👍

AlphaKretin:
I've learned the hard way why a few patterns I didn't like in TectonicData have to be that way, so I've closed those comment threads. As long as the jank is contained to that file and it works smoothly when called elsewhere (which I believe is the case), we're cool. Might be worth adding comments explaining why it is how it is to the file though.

Besides that, only bugbear is the `as unknown`/saving empty `populateMap`s to every entity thing, which I do believe should hopefully be resolvable.

Cincidial:
Sounds good, I'll add some comments there & take a look at the unknown cast
And yes all the jank to make it work smoothly elsewhere should all be in that one file haha

src/app/data/tectonic/TectonicData.ts
Cincidial: I did find a slight improvement to this. Instead of `undefined!` it should be able to just be an empty struct that gets replaced `{}`

src/preload/loadedDataClasses.ts
```diff
@@ -156,7 +176,7 @@ export class LoadedPokemon extends LoadedData<LoadedPokemon> {
     spDefense: number = 0;
     bst: number = 0;
     abilities: string[] = [];
-    levelMoves: [number, string][] = [];
+    levelMoves: LoadedPokemonLevelMove[] = [];
```
Cincidial, line 179: Level moves and forms for some reason required some extra changes because JSON stringify was just dropping the arrays / values if it was empty. This in turn made it so that the `as LoadedDataJson` would break as those fields were not present on **every** entry

src/app/data/abilities/TwoItemAbility.ts
```diff
@@ -27,9 +27,9 @@ export const twoItemAbilities: Record<string, validateItemFunction> = {
 
 export class TwoItemAbility extends Ability {
     private validate: validateItemFunction;
-    constructor(ability: LoadedAbility, validate: validateItemFunction) {
+    constructor(ability: LoadedAbility) {
         super(ability);
-        this.validate = validate;
+        this.validate = twoItemAbilities[this.id];
```
Cincidial, line 32: I kept this change because the value passed in was this anyways. So just removes the need to access the subclass array from outside of this class

Cincidial:
I think that's everything from the first round of reviewing? Let me know if there's more. Thanks!

src/preload/loadedDataClasses.ts
AlphaKretin: Ah yeah it gets annoying with JSONifying mixed arrays like this

src/app/data/abilities/TwoItemAbility.ts
AlphaKretin: Oh, yes, this is how I've done it in all the moves and items, I didn't realise I was still doing it the other way! That explains why you sought to change it. Good eye.

AlphaKretin:
Beautiful, tyvm!

Cincidial:
Anytime!

PR #187, opened 2025-05-08
U/cincidial/#169/on hover abilities and tribes

Cincidial:
Just posting if you want to pull and take a sneak peak. 
While doing the hover stuff I had some ideas for condensing info in the modal.

The commit to add the new components are separate from this modal work. Just adds some capsule displays for abilities and tribes as well as a component for the form changing buttons

WIP
- Move list & details for both those tabs
- Touch ups to matchups tab
- General look and feel

### DIFF
--- a/src/app/data/tectonic/Pokemon.ts
+++ b/src/app/data/tectonic/Pokemon.ts
@@ -47,7 +47,12 @@ export const defaultStylePoints: StylePoints = {
 
 function getterFactory<T extends keyof Pokemon>(mon: Pokemon, key: T) {
     return function (form: number = -1) {
-        if (form in mon.forms && mon.forms[form][key] !== undefined) {
+        if (
+            form in mon.forms &&
+            mon.forms[form].id !== "" &&
+            mon.forms[form][key] !== undefined &&
+            mon.forms[form][key] !== ""
+        ) {
             return mon.forms[form][key];
         }
         return mon[key];
--- a/src/app/pokedex/components/EStatRow.tsx
+++ b/src/app/pokedex/components/EStatRow.tsx
@@ -5,8 +5,8 @@ export default function EStatRow({ name, pokemon, form }: { name: "PEHP" | "SEHP
 
     return (
         <tr key={name} className="italic">
-            <td>{name}</td>
-            <td>{ehpValue}</td>
+            <td className="text-right">{name}</td>
+            <td className="px-3">{ehpValue}</td>
             <td></td>
         </tr>
     );
--- a/src/app/pokedex/components/MoveDisplay.tsx
+++ b/src/app/pokedex/components/MoveDisplay.tsx
@@ -28,7 +28,7 @@ export default function MoveDisplay({
         <>
             <div className="flex">
                 {/* Moves List */}
-                <div className="w-1/2">
+                <div className="w-1/2 overflow-y-auto">
                     {levels.length === 0 && (
                         <div className="mb-2">
                             <label className="text-gray-600 dark:text-gray-300 mr-2">Category:</label>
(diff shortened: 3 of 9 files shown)


public/move_categories/Adaptive.png
Cincidial, line 1: I'm by no means a grapihcs artist lol. This is just the pushed together. Consider it a placeholder?
Also I got these from bulbapedia. I couldn't find a nice source of SVGs like for the types. Lemme know if this is an issue. 

I'm assuming since it's on bulbapedia it's fair game to use?

AlphaKretin:
Excited by the prospect of a UI refresh. Can't take a look rn but will later

public/move_categories/Adaptive.png
AlphaKretin: It should be, but also, we have icons for this in Tectonic, may as well use those instead to be sure and for consistency

AlphaKretin:
My first impression of this is that overall, I like it a lot!

It can seem a bit cramped, but like, obviously that can be a good thing because it means we're not wasting space. Padding can add up quickly.

I'm not 100% sure if the tribes and abilities obviously read as that without headings or other indication. I'm also not sure I like the ability descriptions being exclusive to title text, especially on mobile where "hovering" doesn't really exist.

The stat display could still be a bit prettier, but that's not necessarily in scope here - it has its own section carved out and we can update it separately whenever. Right now it's obviously just how it was, so that's fine.

I like the new move display in spirit, but something I'd been meaning to fix is that some info wasn't display on the old move list - in particular, priority and targeting. In this new table view, I don't know that there'd be room to fit that at all.

Overall only minor nitpicks though, I do like the direction here.

public/move_categories/Adaptive.png
Cincidial: Are there small circular ones?
I know of the rectangle ones, but I was having trouble making them circles and still looking okay.

AlphaKretin: Mm, that's a good point, it wouldn't look great. Using the Bulbapedia ones should be fine then - the usage notes on the image page don't lay claim to the pictures, obviously, since they're from the games.

Cincidial:
Thanks for the feedback!

I agree on the being cramped. As I added more to that first page it started to feel very dense.
Will see if things can be padded a bit / have headers / split into another tab (I like the tabs except for when there's a need to scroll the tab list)

Ah right I always forget mobile doesn't have hovering. I'll do some thinking on it, that's a good point. 
For labeling though I'll put it in a table, the longest section there is 3 tribes which still has room for a label on the left. 

Did you have any ideas for direction on the stat display?

I'll add the new data to the moves table and see how things can fit (or not)

AlphaKretin:
> Did you have any ideas for direction on the stat display?
Not off the top of my head, just a vague sense that it looks vague and clunky. Like I said, doesn't need to be part of this PR.

Having a few more tabs would be good, and would make room for things like ability descriptions without having to worry too much about vertical space.

Cincidial:
@AlphaKretin sorry for the late update, but a summary:
- Add move prio. I didn't do the targeting, I have an idea of how to a get a small one into the table (picture a 2x2 table with `u` for the user and `f` for the foe, then just shade the appropriate boxes) but tbh I don't find that data very useful myself lol so I didn't have a lot of motivation to do it. With just move prio the table space is okay. 
- Fixed a couple bugs with forms data and weight
- Added display data for weight and height incase the pokemon has an ability relevant to those and you are looking up other mons. 
- Moved encounters and evos to a new tab
- Updated the ability and tribes display to have the descriptions in line, table style. 
- I had an idea for the stats display, so I kinda just did it here. Lemme know what you think. It should have a better color scheme and a more standardized width.

src/app/data/tectonic/Pokemon.ts
```diff
@@ -45,12 +45,21 @@ export const defaultStylePoints: StylePoints = {
...
 };
 
-function getterFactory<T extends keyof Pokemon>(mon: Pokemon, key: T) {
-    return function (form: number = -1) {
-        if (form !== 0 && form in mon.forms && mon.forms[form][key] !== undefined && mon.forms[form][key] !== "") {
-            return mon.forms[form][key];
+function getterFactory<T extends keyof Pokemon, R>(
+    mon: Pokemon,
+    key: T,
+    isNull: (value: R) => boolean
```
AlphaKretin, line 51: This is good catch and a great solution, ty!

AlphaKretin:
I like the new stat display! Overall I'm now really happy with the aesthetics of this. I could continue to nitpick (maybe a bit more padding there, maybe rounded corners on the table entries for abilities and tribes), but that doesn't seem necessary. 

As far as move targeting goes, I did have one piece of feedback specifically noting that it was missing. I do like that idea for how to display it, it matches how it is in Tectonic as well. I understand it doesn't fit well in the table, but I'm concerned that we don't list it anywhere at all. This was one advantage of the old "browser" style view the move list used to have, but I also understand the benefits of this new approach. I think we can leave it for now, and make a new to-do list item of a Move Modal that matches the information that can be seen in the MoveDex in-game. That can have all the nitty gritty details, and we could even remove some details from the abridged table.

With that, then, I think we're good? I think we're good. I'll leave it for a bit in case you *do* want to touch up the padding or something, but again no real need, so just lemme know if you think we're good to go too.

src/app/data/tectonic/Pokemon.ts
Cincidial: No problem!
I think I had introduced the bug in the data parsing update - these values became set to defaults instead of undefined as the form was no longer a partial pokemon

Cincidial:
Thanks for all the kind feedback!
Yup I think we're good to go. Can definitely add some follow up issues to work on the nitpicks & move info in those.

Just an idea for the move rows, there could be a toggle button that swaps between the description and other info. Or also the description could be underneath all the info stretching the whole row. That could free up a lot of space for other stuff. I can try some stuff out on another branch (this one's getting a little big for even myself haha)

AlphaKretin:
Oh, if each move had two rows, one with the stats and one with the description, that could be a really good solution! But yeah, as a separate thing.

PR #194, opened 2025-05-14
#181 - Item list work

### LINKED ISSUE #181: Items list in Pokedex
I think the pokdex item list is missing some items. I don't see the super effective berries in there and some other stuff.

Cincidial:
Summary:
- Gets TM data from the tectonic repo via the pre load script
- Adds 2 new cols to the items data table in the pokedex (is held and wild pokemon list)
- Adds ability to filter by the 2 new cols
- Changed items table to show all items instead of just wild held ones
- Changed rows so that only rows with pokemon that hold the item can be clicked to filter the pokemon list
- Changed the filtering to found wild pokemon for the items to use a dictionary so the first load is faster
- Added a few missing pictures for the ones I could find for the items

TODO:
- [x] Add missing item pic
- [x] Filter item wild list to be unique. 
- [x] Revert back to only showing some items by default

Closes #181


public/Items/NORMALMEMORY.png
Cincidial, line 1: The other ones look slightly different. I found their equals to this in the tectonic graphics folder, but the current pics look cool being slightly different from `NORMAL`

src/app/pokedex/page.tsx
```diff
@@ -227,23 +241,59 @@ const Home: NextPage = () => {
...
-                                    </tr>
-                                ))}
+                                {itemDisplayData
+                                    .filter((i) =>
+                                        itemFilter === "Held"
+                                            ? i.item.isHeldItem
+                                            : itemFilter == "Wild"
+                                            ? i.wildMons.length > 0
+                                            : true
+                                    )
```
Cincidial, line 273: Yeah the filtering could definitely be done cleaner (ie. convert the string to methods that perform the filtering and what not), but it's only 2 filters and idk if anymore will be added for this page.

AlphaKretin: We could add name filters and stuff analogous to the Pokemon filtering, but that doesn't need to be part of this PR, and it's easy enough to clean it up to match those patterns later if we decide to.

public/Items/ZYGARDECUBE.png
AlphaKretin, line 1: Tectonic has a lot of items in the data that are just sort of default with Essentials and utterly unused in the game, this included - for clarity of useful information it would be nice to find a coherent way to filter them out, but I doubt there's any sort of consistent logic we could use. I have asked for a TODO on the Tectonic end to remove dummy data like this for our sake.

AlphaKretin:
If an item is held by a Pokemon multiple times (i.e. in multiple rarity slots or across multiple forms), then it'll appear that many times in the list of holding Pokemon, which feels pretty redundant and looks janky. Would be nice to filter those lists to unique.

Other than that, though, looks great!

AlphaKretin:
Actually, one other issue - despite them being added to the data, I'm not seeing TMs in the list here I don't think.

public/Items/ZYGARDECUBE.png
Cincidial: There is a items `cut` file. Not sure if cut actually means gone or something else though. Assuming that it means something else. 

Maybe we could assume only items in pocket 5, held by wild, or TM should be displayed? Could be a good enough solution for now

AlphaKretin: Cut is for items that were in the game, but are now intended to be removed, but still need definitions for the purposes of save data conversion. The unused/dummy items are still in the main file I believe, just... not used.
Restricting the items shown was the solution I was using before, yeah, I'd be happy going back to that for now.

Cincidial:
> If an item is held by a Pokemon multiple times (i.e. in multiple rarity slots or across multiple forms), then it'll appear that many times in the list of holding Pokemon, which feels pretty redundant and looks janky. Would be nice to filter those lists to unique.
> 
> Other than that, though, looks great!

Good idea. Added it to the todo. 
I'll probably get to those items tomorrow (friday for me) / weekend. Works been keeping me pretty busy this week.

src/app/pokedex/page.tsx
```diff
@@ -227,23 +243,69 @@ const Home: NextPage = () => {
...
+                                    .map((i) => (
+                                        <tr
+                                            key={i.item.id}
+                                            onClick={() => (i.wildMons.length > 0 ? handleItemClick(i.item) : () => {})}
+                                            className={`hover:bg-blue-50 dark:hover:bg-blue-900 ${
+                                                i.wildMons.length > 0 ? "cursor-pointer" : "cursor-text"
+                                            }`}
+                                        >
+                                            <TableCell>
+                                                <img
```
Cincidial, line 285: Cannot use `Image` otherwise this turns into an infinite loop when trying to render a missing image

AlphaKretin: Found a pattern that makes it work, see latest commit
It may have also been just because the NOT_FOUND url was *also* erroneous in the format Next was expecting though, it was missing a preceding slash.

AlphaKretin:
Should be good from here but I'll let you check over my tweak before I merge

Cincidial:
> Should be good from here but I'll let you check over my tweak before I merge

I'm good with that. 

Another idea as well as I have a feeling we'll want to do this with other images. 
Should there be a component for this that just wraps `Image` and handles the image missing state itself?

AlphaKretin:
Oh that'd be wise, yeah. 

I've just laid down so I'll merge this when I next get up just because tweaking the commit desc and stuff is a pain on mobile.

PR #200, opened 2025-05-19
Adaptive icon improvment

### LINKED ISSUE #171: Improve Adaptive category icon
Can do in a similar way to the type icons. Add in the physical / special / adaptive / status icons. 
Can be used later for a short way to indicate category of a move.

Cincidial:
Much better than the overlay version, could probably be touched up a bit still though.

For an example look at Espeon for rising power
image

Closes #171


AlphaKretin (approved):
I won't merge right away since you mentioned wanting to touch it up more, but feel free to for now if you want to - better to have this and tweak it again later than stick with the old one.

Cincidial:
Ah, idk what else to do. Just feel like it needs something. 
Might revisit in the future, but will merge for now as it's much easier to see than the first version.

Cincidial:
Ah also, just realized after merging that since this is basically a new image, it's not gunna be transformed till the next time we reset image transforms. So for now on prod this has the alt text lol

PR #207, opened 2025-05-25
Ability table filters for team building

### LINKED ISSUE #206: Pokedex abilities table filters
Add filters for team building related abilities
- Additional type
- Immunities
- Resists
- Weather (Sun, Rain, Hail, Sand, Eclipse, Moonglow)
  - Probably add a new ability subclass for these (1 per?)
  - Actually, instead of subclassess I'm just going to check the flags as it's more future proof and for this usage we don't care about what the ability actually does
```
const abiltiesWeatherFilter = { // Note also AllWeatherSynergy
    Sun: "SunSynergy",
    Rain: "RainstormSynergy",
    Hail: "HailSynergy",
    Sandstorm: "SandSynergy",
    Moonglow: "MoonglowSynergy",
    Eclipse: "EclipseSynergy",
}
```

Cincidial:
Closes #206 

As mentioned in the issue I did not follow the subclassing pattern for the weather ability checks:
- Flags offer a more future proof / less maintenance way of checking 
- We don't care about what the ability does in this case, so subclassing is overkill
- I don't really know how you want the subclassing setup for the calculator for weather abilities, so this way is pretty simple and easy to replace later.

Filters have a focus on stuff that you'd use for team building. Let me know if there's others you'd like / know users would find useful. i just went with stuff I normally would check for when making a team

Techo style filtering used instead of the filter style on the pokemon table page because in this case abilities will always be uniquely filtered - ie. applying 2 filters is basically always going to be overkill and result in nothing. 

Note: I know the filters right below the page "filters" looks kinda odd. I want to update the page selection in another PR to feel more like a header and not be sticky (and instead have the page support a scroll to top floating button)

image
image

### DIFF
--- a/src/app/data/tectonic/Ability.ts
+++ b/src/app/data/tectonic/Ability.ts
@@ -23,5 +23,33 @@ export class Ability {
         return 1;
     }
 
+    public hasSunSynergy() {
+        return this.flags.includes("SunSynergy");
+    }
+
+    public hasRainSynergy() {
+        return this.flags.includes("RainstormSynergy");
+    }
+
+    public hasHailSynergy() {
+        return this.flags.includes("HailSynergy");
+    }
+
+    public hasSandSynergy() {
+        return this.flags.includes("SandSynergy");
+    }
+
+    public hasEclipseSynergy() {
+        return this.flags.includes("EclipseSynergy");
+    }
+
+    public hasMoonglowSynergy() {
+        return this.flags.includes("MoonglowSynergy");
+    }
+
+    public hasAllWeatherSynergy() {
+        return this.flags.includes("AllWeatherSynergy");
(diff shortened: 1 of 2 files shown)


AlphaKretin:
Agree that flags are the way to go here, subclassing is only a relevant pattern for damage calculator functionality as far as I intend.
> Note: I know the filters right below the page "filters" looks kinda odd. I want to update the page selection in another PR to feel more like a header and not be sticky (and instead have the page support a scroll to top floating button)

👍

AlphaKretin (approved):

Cincidial:
Will merge later tonight ~1.5 hours

Cincidial:
Deployed, checked the website and changes look there and things are still running

PR #208, opened 2025-05-25
Type chart attack dual type

### LINKED ISSUE #139: Add attack vs dual-type matchup to the main type chart
On the main type chart add an on-click to the attack (maybe def types too) that replaces the top left corner (attack -> defender text) with the type clicked. The recalc all the table cells as though they are dual types with the clicked type being the move type used. 

Similar to this per-type one here https://pokemondb.net/type/electric
Image

Cincidial:
Closes #139 

Figured this was a good time to clean up usage of `useShort` as it's no longer needed with the icons for types.

Also added in the dual type matchups when clicking a row in the type chart. To clear it you can click the icon in the top left 
or the row again. Might not be the most intuitive thing, but also takes up no additional UI space so it's kinda nice that way.

Tooltips when hovering show both defending types. Can't grab it in a screenshot though.

image
image


Cincidial:
Will merge later tonight ~1.5 hours

Cincidial:
Deployed, checked the website and changes look there and things are still running

PR #212, opened 2025-06-03
Use new move table in Pokedex

### LINKED ISSUE #205: Pokedex move table
Once #204 is merged the main move table for the pokedex should be updated to use that table component so it has all the new info and filtering. 

Check out adding sorting as well here, not required though. 
Make sure to keep the on click for the main table to add a pokemon filter

Cincidial:
Closes #205
Changed pokedex to use new move table. 

Lost some coloring as the background is no longer the move color. It's easy to get it back in but hard to get it to look good as it washes out the type icon and the category icon. Could find another way to add some flare with the move type color, but leaving as is for now

image

### DIFF
--- a/src/app/pokedex/components/MoveTable.tsx
+++ b/src/app/pokedex/components/MoveTable.tsx
@@ -32,7 +32,15 @@ function MoveTargetCell({ move, position, children }: { move: Move; position: bo
     );
 }
 
-export default function MoveTable({ moves, showLevel }: { moves: [number, Move][]; showLevel: boolean }) {
+export default function MoveTable({
+    moves,
+    showLevel,
+    onMoveClick = undefined,
+}: {
+    moves: [number, Move][];
+    showLevel: boolean;
+    onMoveClick?: (m: Move) => void;
+}) {
     const displayableMoveFlags = new Set<string>();
     displayableMoveFlags.add("Sound");
     displayableMoveFlags.add("Punch");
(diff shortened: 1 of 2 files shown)


src/app/pokedex/components/MoveTable.tsx
```diff
@@ -109,7 +121,18 @@ export default function MoveTable({ moves, showLevel }: { moves: [number, Move][
...
+                                <tr
+                                    className={getRowClass(index)}
+                                    onClick={() => onMoveClick?.(m)}
+                                    onMouseOver={(e) => {
+                                        e.currentTarget.className = getRowClass(index, true);
+                                        e.currentTarget.nextElementSibling!.className = getRowClass(index, true);
+                                    }}
+                                    onMouseOut={(e) => {
+                                        e.currentTarget.className = getRowClass(index);
+                                        e.currentTarget.nextElementSibling!.className = getRowClass(index);
```
Cincidial, line 133: Tried doing this with state as it was much more straightforward, but it was horribly laggy. So, a custom on hover was needed - the details and description are 2 separate rows so if you use `hover:` with tailwind it only does "half" of the move.

AlphaKretin: If state tracking is causing lag, one way to potentially alleviate that is to extract the stateful element into its own component, so that only it rerenders when necessary, instead of the entire parent component.

Cincidial:
Closing as there's additional changes I want to do first.

PR #213, opened 2025-06-07
Encounter Tracker

### LINKED ISSUE #198: Feat: Encounter Tracker
I'm not sure how many people nuzlocke the game, but a something like an encounter tracker could be helpful.
As likewise via the dmg calc you can see trainer teams and prep, this tool would help track the other side of things.

Got inspired by this site
https://nuzlocke.app/game
I've been playing a bit of volt white 2 redux and am using that site to track encounters. 

For me personally I have an interest for tectonic because although I probably won't nuzlocke it, I do like restricting what I can catch in the same way as the standard rules (1 per "route", first encounter, dup rule, shiny clause). Otherwise I would just always use my favorites!

Could add an option to allow 1 per encounter table in each area (like each grass + water)

Cincidial:
Closes #198 

**Things to look for in this PR**
- Idea on how to better sort the encounter tables
  - Current approach is the naieve one of just sorting by level ascending and splitting special / regular encounters by a filter.
  - This works okay for the most part, but the start of the game is a little wonky (picking your starter in the special tab is not the first thing)
  - We can manually force things to specific spots, but just wondering if there's a better way. 

Supports
- Saving through local storage for persistence
- Multiple playthroughs via names (uses a key behind the scenes so you can have duplicate names)
- Filtering between regular and special encounters
- Mobile friendly

Does not have extended support for
- Dupes clause 
  - As far as I can tell this isn't really needed for Tectonic
- Shiny clause 
  - Generally this is just a free catch, so not really any tracking support needed
- Randomizer 
  - Can be done in the future by having all the locations but just providing a drop down search for all pokemon instead of the preset options

Future work
- Ranomizer support
- Add pokemon modal on click of the pokemon picture
  - Given how this page is mobile friendly, I'd like work to be done first here to make the modal more mobile friendly (can also make it it's own page)


src/app/data/tectonic/Encounter.ts
Cincidial, line 1: The loaded class for this just had more data and the stuff I needed, where this is basically just a subset. So changed stuff to use the loaded class

PR #216, opened 2025-06-17
Merge when Tectonic 3.3 releases

AlphaKretin:
Setting up for a one-click update


AlphaKretin:
Update is out but not yet merged into main, will happen after waiting to see if any urgent hotfixes are needed

PR #223, opened 2025-06-24
Fixed null check for form stats

### LINKED ISSUE #218: Missing stats for some forms
Image

Cincidial:
Closes #218 

Old issue unrelated to 3.3, the null check for stats not being set for a form would have been checking the actual reference object for `blankStats`, but the values coming from json build a new stat object everytime. So that would never be true for the null check. 

I've opted to just check if `hp == 0` because (in my mind) nothing will ever have 0 hp, so if it is 0 then it must not have been set in the form.


AlphaKretin (approved):
Tysm for taking care of all these issues

Cincidial:
> Tysm for taking care of all these issues

Anytime!

PR #227, opened 2025-06-25
Add proper handling for "Dry Skin" and similar abilities

### LINKED ISSUE #225: Add support for abilities that apply both weaknesses and immunities
Dry Skin, Fine Sugar

AlphaKretin:
Close #225 

A couple of issues:
- ~~There's no handling to convert a "1.25x" damage multiplier to anything nice-looking in the Pokedex. Theoretically if the ability somehow appeared on another Pokemon with different relationships to Fire or Water, that could create other edge cases too. Handling each of them uniquely is probably not feasible, we can leave it decimal, just gotta make it a *bit* nicer looking and colour the background.~~
- These abilities are seemingly have no effect in the damage calculator, only in the Pokedex, despite them calling the same `calcTypeMatchup` function iirc.

### DIFF
--- a/src/app/data/abilities/MatchupModifyAbility.ts
+++ b/src/app/data/abilities/MatchupModifyAbility.ts
@@ -5,7 +5,7 @@ export abstract class MatchupModifyAbility extends Ability {
     protected affectedTypes: PokemonType[] = [];
     protected matchup: number = 1;
 
-    private affectsType(type: PokemonType) {
+    protected affectsType(type: PokemonType) {
         return this.affectedTypes.map((t) => t.id).includes(type.id);
     }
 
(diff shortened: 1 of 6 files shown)


AlphaKretin:
Point 1 resolved by #229

Cincidial:
Oh for the second point I think they have no effect on the dmg calc because they get created as the immunity subclass first. We'd need maybe a new ability subclass that supports both an immunity and weakness at the same time?

Cincidial:
> That's meant to be exactly what this PR was changing, and if it was still the issue then at least the immunity should be reflected in the damage calculator. I wonder if it's an issue of using the parent class's modifiedMatchup function instead of the subclass's...

Whoops, I didn't look at the code haha. 
From looking at the code I don't really see what the issue would be, looks like the standard setup.

AlphaKretin:
Right? I'll try again tomorrow, see if it's suddenly just started working, maybe something weird and one off was going on last time.

AlphaKretin:
With that silly, silly bug resolved, this now works fine. Yay.

Cincidial:
> With that silly, silly bug resolved, this now works fine. Yay.

What was the fix?

PR #243, opened 2025-07-06
Swap images to html img tag and src github

Cincidial:
Sourced to main branch in the repo, makes it hard to dev but that can be touched up later. 
Locally this all loaded (from github). Though it's noticeably slow on the first load. It loads though

### DIFF
--- a/src/app/pokedex/page.tsx
+++ b/src/app/pokedex/page.tsx
@@ -359,7 +359,7 @@ const Home: NextPage = () => {
                     </table>
                 </TabContent>
                 <TabContent tab="Items" activeTab={activeTab}>
-                    <div className="overflow-x-auto">
+                    <div className="overflow-x-auto mx-auto">
                         <table>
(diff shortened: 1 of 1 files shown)


Cincidial:
Results in a build warning complaining about using `img`. Something we will just need to live with.

Also, this will need a followup to help make local dev easier. Cause it basically forces looking at main & will stop working if you have no internet (idk if that's a problem, but in case you like to dev locally sometimes without access)

Cincidial:
hehe, now we see if we hit the edge requests max and what happens then. Hopefully those were basically all for images though and this should help it

AlphaKretin:
> Results in a build warning complaining about using `img`. Something we will just need to live with.

Yeah this is Next.JS trying to tell us to use its image optimisations... which obviously the point is to avoid, for us. 
In the near future, maybe we could try and find a proper CDN rather than hotlinking to GitHub.

PR #244, opened 2025-07-06
Update to use Sirv CDN

Cincidial:
Updated readme and added Sirv CDN for images. 
Left in github as a fallback incase we run out of space - github can then work until we get rate limited out


AlphaKretin (approved):

AlphaKretin:
What a headache this has all been. I hope things go smoothly from now on.

Cincidial:
Also, expiration header looks good. So caching should last a month for users
image

PR #245, opened 2025-07-07
Local dev use local nextjs for images

Cincidial:
Just so contributors don't have to worry about accidentally taking up CDN limits
Works for `npm run dev`. If you do a `npm start` then you'll be hitting the CDN

Also provides pattern of what it would look like to use nextjs (like we were before) as another fallback for images if the CDNs are out.

Sorry for PR spam on this, but really wanted to get this in so I don't accidentally take up a bunch of our CDN space doing random stuff!


AlphaKretin (approved):
Smart thinking

PR #253, opened 2025-07-11
Trainer items data load fix and fix for forms being -1

### LINKED ISSUE #241: Held items reportedly missing on documented trainer teams
> I've seen multiple missing items from trainers
There was definitely a missing strike vest on the first trainer of cave of hatching - I remember this as I learned about it the hard way
Also I'm pretty sure there was at least one missing item (probably more) in the tournament though I can't remember exactly who or what

From discord. To be investigated.

Cincidial:
Closes #241 
Pls pull locally for a quick data check as I updated the loadedData.json tyty


src/app/data/teamExport.ts
```diff
@@ -219,7 +219,7 @@ const decodeChunk = (
 
     mon.species = loadedMon;
     const formIndex = loadedMon.forms.findIndex((f) => f.formId === formId);
-    mon.form = formIndex;
+    mon.form = Math.max(formIndex, 0);
```
Cincidial, line 222: With the new pokemon card in the team builder, if there was a form to change to the first click of the arrows would do nothing (as it would go from -1 back to 0) then it would work as normal. This fixes that.

PR #254, opened 2025-07-11
Solar Calc

### LINKED ISSUE #203: "Hex" defaults to double damage regardless of Status field
It will correctly update if *changed* to None, but not properly check when it first loads. Possibly affects other conditional double moves too.

Cincidial:
Closes #203 

Features
- Search by minidex for both player and opponent
- Team code and saved team load for player
  - This includes being able to export and save the team with any changes (just incase you are building a custom team for a fight)
- Trainer search in a similar fashion to the minidex filter
  - After selecting a pokemon from a trainer I still need to disable editing of the mon. **This is done, but you need to click off the mon and back on to reload if it gets changed. The data is not saved to the trainer though**
- Move calc display for all player and opponent moves at the top
  - This is currently quite ugly, but functionally everything should be there. Making this pretty is gunna take some time as it's really info dense, but displaying the data in as small a profile as possible is optimal... Modal is out of the question, but can maybe include a more details for stuff by clicking on it. Preferably though everything is visible at a glance. 
- Field effects per side
- Battle effects affecting both sides

TODO:
- [x] Show adaptive category after taking into account user stats
- [x] Other move data (bp, acc, prio, basically look like move table row?). I just tossed this info on the move card, it doesn't look good though. However, I think figuring out a good UI for this is going to take some time and I'd rather just display the stuff as is for now then pretty it up later
- [x] Make crit not rely on the same component being rendered (move it into the passed in prop data via moveData like before)
- [x] Show all calc data
- [x] Custom Var things (honestly Idk what's going on with this, ima just try to get it like how it was before)
- [x] Update with main. Stat stepper left as +/- inputs so that negative numbers can be input and an assert in the calculator can be avoided
- [x] Moves under card?
- [x] Dmg color based on hits to KO
- [x] Indicate who moves first


Cincidial:
Finally got around to doing some more work on this. 
It's coming along. Can now see the damage calc. Just gotta to some touch app and re-add stuff that I removed to reduce compilation fixes to build
image

(Added to the main post the todo)

AlphaKretin:
holy moly

Cincidial:
For move prio I'm listing along side the move, as you don't pick a specific move to use. It Just shows the calc for all 4 at once

AlphaKretin:
Oh, right, yeah, good point

AlphaKretin:
I still need to look over the code but I've used this in practice and it feels great, kudos!

AlphaKretin:
image
Actually I guess one minor point, having the accuracy in the text box here feels a bit weird at times, my eyes often go to it when I'm trying to look for the damage percentage. The priority on fakeout it's also unclear what that is

src/app/damagecalc/components/SideStateUI.tsx
```diff
@@ -0,0 +1,31 @@
+import { nullSideState, SideState } from "@/app/data/battleState";
+import Checkbox from "@/components/Checkbox";
+import { ReactNode, useState } from "react";
+
+const sideStateNameMap: Record<keyof SideState, string> = {
```
AlphaKretin, line 5: Good job splitting this off to account for the new symmetry of the calculator!

src/app/damagecalc/page.tsx
```diff
@@ -1,203 +1,86 @@
...
-            return filters.every((filter) => {
-                return filter.apply(mon, filter.value);
-            });
-        });
-        return filtered;
-    }, [filters, mons]);
-
-    function updatePlayerPokemon(card: Partial<PartyPokemon>) {
-        setPlayerPokemon(new PartyPokemon({ ...playerPokemon, ...card }));
+    const [gravity, setGravity] = useState<boolean>(false);
```
AlphaKretin, line 39: If we implement more field conditions in the future - which I think there will be - we might want a bundled system like we have for SideStates, rather than each one getting its own thing like this.

src/app/damagecalc/page.tsx
```diff
@@ -1,203 +1,86 @@
...
+    const [multiBattle, setMultiBattle] = useState<boolean>(false);
+    const [playerSideState, setPlayerSideState] = useState<SideState>(nullSideState);
+    const [opponentSideState, setOpponentSideState] = useState<SideState>(nullSideState);
+    const [playerMoveData, setPlayerMoveData] = useState<MoveData[]>([]);
+    const [opponentMoveData, setOpponentMoveData] = useState<MoveData[]>([]);
+
+    const matchingTrainers = Object.values(TectonicData.trainers)
```
AlphaKretin, line 46: In the future I want to sort trainers by rough in-game or pokemon level order, rather than alphabetically - if only so that the endgame AP fights don't immediately show up - but that's a future thing, probably part of #262

src/components/PokemonCardHorizontal.tsx
```diff
@@ -1,16 +1,27 @@
...
+import {
+    MAX_LEVEL,
+    MAX_SP,
+    MAX_STEP,
+    MIN_LEVEL,
+    MIN_SP,
+    MIN_STEP,
+    STYLE_POINT_CAP,
+    styleFromStat,
+} from "@/app/data/teamExport";
```
AlphaKretin, line 16: These probably still need a better home than the teamExport file lol, they've been awkwardly stranded there for a while despite having a bunch of uses unrelated to that.

src/components/TypeChartCell.tsx
```diff
@@ -25,10 +25,10 @@ function getColourClassForMult(mult: number): string {
         return "bg-not-very-effective";
     }
     // if neutrally effective, no special colour
```
AlphaKretin, line 27: This comment is a bit outdated with this new functionality

AlphaKretin (approved):
I've pointed out a bunch of nitpicks, but I don't think any of them are blocking - I'd rather get this big change in people's hands sooner and iron out the details later. Up to you though, if you want to tackle some of this, I won't merge it right away, but feel free to do so yourself if you feel it's good to

src/app/damagecalc/page.tsx
AlphaKretin: I added new PBS data for encounter availability order (in terms of level caps), idk of any such plans for trainers.

Cincidial: OH yeah that is actually a great idea sorting like that. Super convenient just to see it right at the top / scroll to it

Cincidial: Whenever that PBS encounter data is available in github I'll add it in to be used by the tracker. Just make an issue and assign it to me whenever!

src/app/damagecalc/page.tsx
```diff
@@ -207,224 +90,287 @@ const PokemonDamageCalculator: NextPage = () => {
...
-                                    {isReadyToCalculate() ? (
-                                        <DamageResultComponent damageResult={damageResult} />
-                                    ) : (
-                                        <div className="text-center text-gray-500 p-8">
-                                            <p>Select Pokémon and moves to see damage calculation</p>
-                                        </div>
+                        <hr className="w-full my-3" />
+                        <div className="flex gap-2">
+                            <FilterOptionButton
+                                isSelected={showTeamSearch}
```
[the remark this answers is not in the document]
Cincidial: For this one I'd say that makes sense, but leave it till the next time we need to add one and do it then. 
Since both sides need a separate state to track them, we'd still need 2 state variables. There's only 2 each right now so nothing crazy.

src/app/damagecalc/page.tsx
AlphaKretin: The data is already on dev, so went ahead and did so, more details in the issue ofc

src/app/data/types/PartyPokemon.ts
```diff
@@ -23,18 +23,21 @@ export class PartyPokemon {
...
+
     constructor(data?: Partial<PartyPokemon>) {
         this.species = data?.species || Pokemon.NULL;
         this.moves = data?.moves || Array(4).fill(Move.NULL);
         this.ability = data?.ability || data?.species?.abilities?.find((a) => a && a != Ability.NULL) || Ability.NULL;
         this.items = data?.items || Array(2).fill(Item.NULL);
         this.itemType = data?.itemType || TectonicData.types["NORMAL"];
         this.form = data?.form || Pokemon.NULL.formId;
-        this.level = data?.level || MAX_LEVEL;
+        this.level = data?.level == 0 ? 0 : data?.level || MAX_LEVEL;
```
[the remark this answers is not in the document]
AlphaKretin: I see. I think I had it knowingly defaulting to level 70, since 0 is an invalid level, but that's pretty inconvenient if someone is trying to backspace and retype a level, so this behaviour makes sense 👍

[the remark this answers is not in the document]
AlphaKretin: > Nevermind, doing the splat means it loses all the methods that are in `PartyPokemon` because it's no longer of that actual type (just has the data).

js is a pain sometimes huh

src/components/PokemonCardHorizontal.tsx
Cincidial: Haha yeah, it's been here for a long time. 
Probably need to a dedicated cleanup up of a couple of this spots at some point, but I'm gunna leave that for later.

src/components/PokemonCardHorizontal.tsx
```diff
@@ -35,14 +46,22 @@ export default function PokemonCardHorizontal({
...
     onUpdate: () => void;
+    showBattleConfig?: boolean;
 }) {
     const [modalMon, setModalMon] = useState<Pokemon | null>(null);
     const [moveIndex, setMoveIndex] = useState<number | null>(null);
     const [stylePreset, setStylePreset] = useState<string>("");
+    const [statusEffect, setStatusEffect] = useState<StatusEffect>(partyMon.statusEffect ?? "None");
+
+    useEffect(() => {
+        // For some reason the status dropdown just refuses to update when changing mons unless it's wrapped in state and a useEffect
```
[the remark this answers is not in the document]
Cincidial: Good point, I did not test hex yet. Will take a quick look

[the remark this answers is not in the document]
Cincidial: Alrighty it kinda works lol. 
For some reason the move initally has the max hex dmg, but after changing the status to something then back to nothing it's fine. 

I'll take a quick look to see if this a quick fix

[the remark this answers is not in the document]
AlphaKretin: The fact it defaults to the max damage regardless of status is the current issue I was wondering if it was fixed - if it's not, that's not this PR's fault (though I mean if you can find a quick fix I'm certainly not gonna say no)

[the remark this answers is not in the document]
Cincidial: Done, initial value was undefined for the party pokemon data. So in the dmg calc the check against `None` would return true for undefined. Fixed it to coalesce to `None` when undefined

src/app/damagecalc/page.tsx
Cincidial: Gunna leave this one for future field conditions to do.

src/components/PokemonCardHorizontal.tsx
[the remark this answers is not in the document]
AlphaKretin: bless. This should now close #203 then

Cincidial:
Fixed several of the nits, but as mentioned others are left for later as separate follow ups / to attach to other work

Cincidial:
Actually, I don't really like the solution for hex. I'm going to update the member variable to no longer be nullable. Cause there's other moves that fall to the same issue. Like facade

Cincidial:
Alrighty, should be good to go from my end. Feel to merge whenever!

AlphaKretin:
Looking great! We can maybe vary the colour of the speed tier chip based on outcome (blue=first, red=last, yellow=tie)? Doesn't have to be now tho

Cincidial:
> Looking great! We can maybe vary the colour of the speed tier chip based on outcome (blue=first, red=last, yellow=tie)? Doesn't have to be now tho

Good idea, I'll add that to the followup issue #271

PR #256, opened 2025-07-18
Added sorting to headers for move table

Cincidial:
Just be aware that sorting in the pokedex move table is quite slow, as expected with re-rendering that entire table lol

- When ascending there's a bar over top of the header
- When descending there's a bar below the header
- Clicking the move search resets to no sort (default order the move list came in)


AlphaKretin (approved):
Looks good! If performance continues to be an issue with our big lists, we might have to look into pagination... but that would be especially awkward for moves
FWIW, on my end, it actually performed fine, though obviously performance is gonna depend on the user's rig, and we want a webpage to be lithe and portable

PR #257, opened 2025-07-18
Add URL parameters for individual Pokedex entries

AlphaKretin:
Need to test this in preview, since in dev it hitches a lot because of how hot-reloading works.


AlphaKretin:
it's been too long i forgot to do a test build to check for errors like this lol

AlphaKretin:
Seems to be working better as I hoped

PR #265, opened 2025-07-20
Proper fraction for 1.25

Cincidial:
Honestly not even sure if it looks better. I tried 5/4 as well, but that doesn't look super good
image


Cincidial:
Gunna close this for now, I think just leaving the decimals looks good. 
These ones are not very common anyways

PR #266, opened 2025-07-20
Added steam power to resisted abilities list

Cincidial:
Steam power was missing, nows properly shows only double dmg instead of x4


AlphaKretin (approved):
Every missing ability we catch is a win, especially when they're easy to add!

PR #267, opened 2025-07-20
Add Vercel as img source

### LINKED ISSUE #237: Vercel rate limits
In addition to the established issue with Image Optimisations, it seems we may soon be running into another rate limit, which sounds like it might have more dire consequences

Image

On the plus side, it's evidence that the website is indeed getting more traffic since the update! But maybe more than it can handle...

Image

Cincidial:
I'm gunna say this closes #237 . Can make another one if it's needed (I really sure do hope it's not)

Adds Vercel as the main image source to take advantage of `<Image` and our 5000 image transforms. The fallback system works as follows
- Try Vercel
- Try Sirv
  - Looking at our stats 2/3 of the way through the month we are over 3GB... We only get 2 per month (after first month) lol. Hopefully with vercel support & it handling populating some users local caches we will be able to support everyone
- Try Github
- Rip

Also added lazy loading for `<img`. This should reduce the load on Sirv a bit if it comes to it. Also speeds up page load when relying on Sirv. Also also helps with the fallback system as doing a request to a new source after falling back will naturally take more time (so lazy loading it should help).

I tested with a single bulbasaur img so as to not spam Sirv. I'm not 100% sure what the error vercel gives when rate limited, but the fallback system worked for when images were not found. The error handler is for any error, so I'm assuming it'll work the same for a rate limit return.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Live at https://tectonic.alphakretin.com
 ## Contributor
 
 If you contribute new/changes to images they'll need to be uploaded to the CDN (Sirv) to be available.
-Images should also be added to the repo for safe keeping
+Images should also be added to the repo as well both for safe keeping and since it's for both local dev and a backup for serving
 Note that Sirv has been configured for a 30 day local browser cache time. So if an image gets updated it may be delayed for users unless they force a refresh.
 
 -   Local dev should always use the localhost nextjs server, unless you use a production version locally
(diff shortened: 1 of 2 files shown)


AlphaKretin (approved):
Clever, clever, clever! This one is hard to test in the dev environment, obviously, because of the different behaviour, so rather than test I'm gonna say this looks good in concept and in code.

PR #268, opened 2025-07-20
SP And Level Input Improvements

### LINKED ISSUE #246: Add presets to Style Point input
Match the presets available in the game

### LINKED ISSUE #264: Optionally allow bypassing SP cap
Allows for comparing offensive and defensive spreads simultaneously without having to switch stats back and forth

Cincidial:
Closes #246 
Closes #264 

SP and Levels now
- Can be free entered only (no +/- input box)
- Removing the last value via backspace auto sets to 0
- Typing a new number when it's 0 will clear the 0 value and set only the number
- Can exceed the min/max values. In practice this mostly just means the max because you can't enter a negative value (since `-` isn't a digit)
- Text becomes red when not in the proper range
- SP presets implemented via drop down selection
  - They also are automatically chosen if you manually enter the values

It's a bit hard to describe what this feels like now. Easiest to just pull and try yourself. 
Tabbing through the fields is pretty quick though for settings values
image

### DIFF
--- a/src/app/globals.css
+++ b/src/app/globals.css
@@ -41,6 +41,7 @@ body {
 
 @theme {
     --color-yellow-highlight: #ffd166;
+    --color-value-out-of-range: #fb2c36;
 
     --color-fire: #e4613e;
     --color-fire-bg: #8a211c;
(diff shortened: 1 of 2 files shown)


AlphaKretin (approved):
Super smooth!

PR #272, opened 2025-07-27
Move Card Followup

### LINKED ISSUE #271: Clean up the move card in the Damage calc
After the Solar update to the calc the move card was left in a purely functional state. It's not very pretty, but that info is there. 
Fix it up!

Cincidial:
Closes #271 

image

### DIFF
--- a/src/app/damagecalc/components/MoveCard.tsx
+++ b/src/app/damagecalc/components/MoveCard.tsx
@@ -100,21 +100,20 @@ export default function MoveCard(props: MoveCardProps): ReactNode {
         return <span>Input for type {customInput === "number"} not yet implemented.</span>;
     }
 
-    // TODO: Make this thing not so ugly, all the data needed to be shown should be here though
     return (
-        <div className="flex flex-col">
-            <div className="flex items-center gap-2 pr-1 py-1 my-1  bg-gray-700 rounded-2xl">
-                <div className="flex flex-col gap-2">
-                    <div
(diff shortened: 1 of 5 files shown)


AlphaKretin:
This looks good, but the damage value and percentage blends in with the other data a bit - maybe it could be bolded or otherwise emphasised?

Cincidial:
Updated to wrap the dmg calc column in a capsule the same way it's done in the header. Updated the screenshot in the original post

Cincidial:
No worries! Take your time (:

AlphaKretin:
Sorry to keep nitpicking after I already said it looks great, but I just realised that effectiveness is only communicated by colour, which isn't great for accessibility. There might be room for text in the header now that all the data is in the bubble below?

Cincidial:
No worries.
That text might be too long, like `Not Very Effective`. 
We don't happen to have icons for this or a commonly understood abbreviation do we? HE, SE, N, NVE, BE (kinda feel like that might be hard to understand)?

I'll see what I can do

Cincidial:
Wait, I'm kind of a dummy. 
What about just using the regular fraction symbol (or text in those special cases). Like `(2x)` and then no text for neutral?
Or maybe that will make the player think the damage is going to get multiplied....

Cincidial:
Note: as another followup at some point, either reduce the text length for some special moves or find a better spot for the custom input var thing. Some moves like earthquake have really long text.

PR #298, opened 2025-09-05
Evo filter in minidex

### LINKED ISSUE #294: Add "fully evolved" toggle to mini filter
Especially for the team builder, where often you're looking at only the end state of a team, being able to cut down the list of mons to scroll through by filtering out NFEs would be useful

Cincidial:
Added the filter for any/first/final evos. Anything with no evolution will be in any filter (as it is both its own first and final evo)

image

Closes #294

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5906,8 +5906,15 @@
             }
         },
         "node_modules/tectonic-damage-calc": {
-            "resolved": "",
-            "link": true
+            "version": "0.1.0",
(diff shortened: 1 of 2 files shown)


Cincidial:
> Thanks for the PR! It might be a bit until I have the energy to take a look due to Life Circumstances

No worries! Whenever it happens it happens. Hope all is well.

AlphaKretin (approved):

PR #299, opened 2025-09-05
Updated img fallback to handle missing images while on dev

### LINKED ISSUE #269: Image loading issues in dev environment
Went to try out the new damage calculator, both for the sake of reviewing the PR and actual in-game use. The page loaded fine at first, but upon scrolling down the list of trainers, I got an error related to the image loading fallbacks. It seems to be an issue with how we handle missing images - ARTIFICALTRAINER is a new class added in 3.3 that we didn't get around to including.

Image

Image

Cincidial:
I scrolled down the trainer list and got `?` instead of crashing. The console also had 400s for things like "ARTIFICALTRAINER". So should be fixed

Closes #269


AlphaKretin (approved):
Works great in my testing too

PR #300, opened 2025-09-05
Changed CV to use real types intead of the full type list

### LINKED ISSUE #295: Restrict Crystal Veil to only real types
Currently shows types like ??? and Mutant as options which is not game-accurate

Cincidial:
image

Closes #295


AlphaKretin (approved):
love a nice and easy sub-one-line fix

PR #301, opened 2025-09-06
Fix moves dissapearing in saved teams

### LINKED ISSUE #275: Certain moves disappearing when saving and loading a team
Reported examples: Accelerock, Acrobatics, Adamantine Press. I assume their lexical proximity is not a coincidence.

Cincidial:
The conditional check to see if a mon had a move in it's IndexMap when saving was resolving to false when the index was 0. It should have only been false when the move was not found (undefined).

Closes #275


AlphaKretin:
Weakly typed language moment

So the reason every missing move was so early in the alphabet was because the issue was with the first move in  moveset. Great catch!

AlphaKretin (approved):

PR #313, opened 2025-10-25
#312 - Fix tribe inheritance

### LINKED ISSUE #312: Issues with Tribe inheritance
Milcery's tribes are FORTUNE,ANIMATED, Blasfemmie's are FORTUNE,SCOURGE. In Tectonic, the two lists are merged, in the website the latter overwrites. I think this is the only case of a weird Tribe inconsistency in the PBS data.

Cincidial:
**Note: This required regen of the data, I'm not sure if this would have broken any data gen you had done for dev/live. Pls check**

**Note 2: If something else that also required data gen was recently merged, this branch will need to be pulled, updated with main, then data gen'd again**

Fixes the case with Milcrey and Blasfemmie
I just spot checked some other pokemon and their tribes look okay still

image

Closes #312


AlphaKretin (approved):
I'll wait for clarification on that note before merging just in case, but otherwise this looks fine

PR #328, opened 2025-10-30
implement type boost abilities

### LINKED ISSUE #286: Implement SCALDINGSMOKE

### LINKED ISSUE #290: TUNNELMAKER causing NaN results

caitlynslashai:
Closes #286
Closes #290 

implements #286 as well as other type-boost abilities 

Steely Spirit, Toxic Attitude, and Verdant do not have their doubles effects implemented

my search was for "Powers up"; any type-based ability with inconsistent wording will have been missed

BaseDamageBoostAbility is getting large; this may or may not be an issue; not my prerogative to answer that


caitlynslashai:
does not fix #290

caitlynslashai:
updated to match previous implementations of type-boosting abilities, which means they're now found in AttackMultBoostAbility. i also added the Regi abilities

this means that this now fixes #290

AlphaKretin:
> Steely Spirit, Toxic Attitude, and Verdant do not have their doubles effects implemented

That's fine, abilities like this need a whole new element added to the damage calc to handle team-wide effects. 

> BaseDamageBoostAbility is getting large; this may or may not be an issue; not my prerogative to answer that

I don't particularly mind it getting large, as long as it's correct that everything in it *is* a base damage boost, this can be checked against Tectonic's code to see which part of the attack calc the ability affects. In this case it looks like you've accounted for that.

I'm gonna edit your OP a bit just to properly link in the issues

PR #332, opened 2025-10-30
partial implementation of fragile locket

### LINKED ISSUE #292: Implement FRAGILELOCKET
At least the stat reduction, if the ability thing is too hard

caitlynslashai:
partially fixes #292 - implements stat reduction but does not implement abilities, which is worth implementing and possible to implement but like. harder


AlphaKretin:
I'll leave this hanging for now and come in to add the ability stuff when I get a chance. It's probably gonna require some special casing.

AlphaKretin:
I'm a big dumb dumb and implemented the ability stuff in a separate PR, #369, forgetting about this. If this fully implements the stat reduction it can be considered ready and the two PRs together will get the job done.

AlphaKretin:
I might see if I can rebase it onto the fragile locket branch so everything's in one place

Issue #184, opened 2025-05-05
Refactor STAB calculation


AlphaKretin:
The current isSTAB method disregards all dynamic type possibilities - forms, abilities, items, etc. It needs to take in a PartyPokemon instead of a species and check all of those things.

AlphaKretin:
Notably this is flagged Pokedex and not Calculator because the isSTAB method is *not* used for calculating STAB damage (maybe it should be once this is fixed?), only the highlighting in the Pokedex and PokemonCard.

AlphaKretin:
Relevant to #281 - oops

PR #334, opened 2025-10-30
281 crystal veil

### LINKED ISSUE #281: Crystal Veil does not affect STAB determination
the `Move#isSTAB` method checks the `Pokemon`'s base types and not the `PartyPokemon`'s potentially modified types. However, this is not for no reason - `PartyPokemon` are not used in the context of the Pokedex view. Trying to modify the function to adapt to both use cases seems to lead to a circular dependency, which upon reflection makes sense - making `Move` require `PartyPokemon` where `PartyPokemon` already requires `Move`s. As such, there's going to need to be some sort of refactor to fix this issuel.

caitlynslashai:
resolves #281 and #184 

i'm marking this as a draft even though it's functional as-is - it breaks the PokemonCard, specifically, but that class doesn't seem to be used anywhere


AlphaKretin:
I don't usually bother approving test deployments because it's a lot of extra activity on both automation's end and mine, but it seems worthwhile here to verify that things are fine after removing that.

PR #340, opened 2025-10-31
252 gravity

### LINKED ISSUE #252: Gravity not nullifying ground immunities

caitlynslashai:
resolves #252 

had to add a second function - but it's clear that there's a need for a dynamic type matchup as well as a non-dynamic one. the only current difference is gravity implementation


AlphaKretin:
I see the use for the separate function, distinguishing between static type interactions for the team builder and dynamic type interactions for the damage calculator. There are a couple of points to question
- Rather than duplicate the function, would it be possible to make battle state an optional parameter to calcTypeMatchup and run the dynamic portions if it's defined?
- Are there any other dynamic elements we might be missing? My brain wanted to say things like abilities or crystal veil, but those *would* appear in the team builder (though Crystal Veil wouldn't appear in the Pokedex, for what that's worth).

AlphaKretin:
Not gonna make the same mistake I did with the FLEX PR - gonna close this and start over, since the last comment points to considering a significantly different implementation.

PR #347, opened 2025-11-01
item select load fixed on dev

### LINKED ISSUE #309: On dev, the item selector shows only TMs instead of excluding TMs
Image

caitlynslashai:
resolves #309 

should be checked to make sure it didn't break things immediately after deployment


caitlynslashai:
note that this does not fix the pokedex view=

PR #366, opened 2026-02-03
Handle new dev-exclusive encounter format in preload

AlphaKretin:
Has two fields for encounter table level instead of just one, to handle sequence breaking vs normal gameplay. Encounter tracker could also be updated to take advantage of this data but that can be a followup

### DIFF
--- a/src/preload/loadedDataClasses.ts
+++ b/src/preload/loadedDataClasses.ts
@@ -432,7 +432,7 @@ export class LoadedEncounter {
 
 export class LoadedEncounterTable {
     type: string;
-    encounterRate?: number;
+    encounterRate?: number | null; // undefined while building json, but becomes null when loaded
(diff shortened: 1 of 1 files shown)


src/preload/loadedDataClasses.ts
```diff
@@ -433,19 +433,30 @@ export class LoadedEncounter {
 export class LoadedEncounterTable {
     type: string;
     encounterRate?: number;
-    levelCap: number = 0;
```
Cincidial: Btw you'll probably need to update the usages of this field in `tracker/page.tsx`

I'm assuming for now just set them all to be `normalLevelCap`

AlphaKretin: OK conveniently it looks like it's already directed through the `EncounterDisplayData` class so we just have to change it there

PR #367, opened 2026-02-03
Fix for checking the version bits

Cincidial:
Very quickly
- Old format stored tectonic version as a u16 **big endian**
- New decoder checks to see if the major of that version has a non-0 value, if so it knows it's an old code
- The major is stored at the first 5bits
- In big endian for a u16 those first 5 bits are actually the second byte, not the first... whoops
- Somehow, by sheer luck, @AlphaKretin the 3.4.0-dev version puts the patch for of `0` (ie. 0 5 bits) along side the `100` zeroes from `4` right where the major was, creating 5 bits of 0

### DIFF
--- a/src/app/data/pokeparty.ts
+++ b/src/app/data/pokeparty.ts
@@ -110,8 +110,8 @@ export class PokePartyEncoding {
         versionU16 |= (parseInt(versionSplit[2]) & 0x1f) << VERSION_PATCH_SHIFT;
 
         const data: [number, BytesRequired][] = [
-            [pokePartyEncodingU8, BytesRequired.U8],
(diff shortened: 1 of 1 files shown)


src/app/data/pokeparty.ts
```diff
@@ -185,7 +185,12 @@ export class PokePartyEncoding {
         const party: PartyPokemon[] = [];
 
         if (view.byteLength >= VERSION_BYTES) {
-            if ((view.getUint8(0) & OLD_CODE_CHECK_MASK) != 0) {
+            decodeTeam(base64);
```
Cincidial: Whoops, gotta remove this

PR #368, opened 2026-02-04
Export team to full tectonic code - downloaded as a file

### LINKED ISSUE #365: Export team to full code for dev import
Would be handy to be able to export a team to the `PokePartyEncodingType.Full` value and have it download as a file for devs to load in to tectonic.

Since this is dev-only the UI for this should not be visible unless it's on a dev version

Cincidial:
Closes #365 

Downloads the team as a file. Done with a new button component so that we can extend this later to PBS downloads with some groundwork complete.

Pls pls pls double check that this **only shows when the data is a dev version**. I checked locally ofc, only saw the button when my loaded json data was for dev.

Only thing I've noticed that might be kinda annoying (not really a way to work around it though) is that download the file again will of course download it to the users downloads location again. Where they may already have this file downloaded - in which case it'll do the ol' append an `(x)` to the file name for whatever x number of download this is.

### DIFF
--- a/src/app/data/tectonic/TectonicData.ts
+++ b/src/app/data/tectonic/TectonicData.ts
@@ -144,6 +144,7 @@ function fromLoadedArray<L extends LoadedData<L>, T>(load: Record<string, L[]>, 
 
 type TectonicDataType = {
     version: string;
+    isDev: boolean;
     types: Record<string, PokemonType>;
     tribes: Record<string, Tribe>;
     abilities: Record<string, Ability>;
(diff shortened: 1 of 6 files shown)


AlphaKretin:
> Only thing I've noticed that might be kinda annoying (not really a way to work around it though) is that download the file again will of course download it to the users downloads location again. Where they may already have this file downloaded - in which case it'll do the ol' append an (x) to the file name for whatever x number of download this is.

Is it possible to make it "Save As" instead of going straight to downloads? Or is this behaviour defined in the user's browser setting?

src/app/data/tectonic/TectonicData.ts
```diff
@@ -217,7 +219,7 @@ Trainer.NULL = new Trainer();
 
 // Start of post-load population
 Object.entries(TectonicData.forms).forEach(([k, v]) => TectonicData.pokemon[k].addForms([Pokemon.NULL, ...v]));
-if (TectonicData.version.includes("-dev")) {
+if (TectonicData.isDev) {
     TectonicData.heldItems = Object.values(TectonicData.items).filter((x) => x.pocket >= 9 && x.pocket <= 13);
```
AlphaKretin: How important is filtering for held items specifically now that we're not using an index-based approach? Not something relevant to this PR but maybe something to think about later.

PR #377, opened 2026-02-08
Implement a bunch of moves/abilities/items

### LINKED ISSUE #83: Implement "Disguise" ability
- [ ] If target's ability is disguise, add a checkbox(?) to indicate whether the disguise is active
- [ ] If it is, pass that information to the calculation function
- [ ] If it is, the damage calculation function returns 0(?)

### LINKED ISSUE #110: Implement fixed-damage moves
Moves like Bide, Endeavor, and Super Fang should bypass damage calculation and do a certain amount of damage directly. Notably, some like Bide would need a custom input, whereas other examples derive their damage from the game state.

Moves in the spreadsheet skipped so far:
- Bide
- Endeavor
- Super Fang
- Counter
- Metal Burst
- Mirror Coat
- Nature's Madness
- Mouthful

### LINKED ISSUE #141: Implement guaranteed crits
When a crit is guaranteed (due to the effect of the move, the opponent being jinxed, etc?), lock the critical hit checkbox and force its value to true. This may be non trivial due to the isolated structure of the MoveCard.

Moves affected up to current spreadsheet progress:
- Storm Throw
- Surging Strikes
- Frost Breath
- Wicked Blow
- Whip Trance
- Critical Error
- Incision
- Humble (Conditional)

### LINKED ISSUE #144: Implement Protect and bypasses
Currently Protect is not implemented in the damage calculator, because it's fairly obvious that if the opponent protects, you'll do no damage, unless using a move that bypasses protect.

However, Empowered moves from avatars can partially bypass protect, doing reduced damage instead of none. This would warrant adding a protect state, and from there moves that bypass protect should read as doing so.

I also have a vague idea that there are other moves that are only partial protections, like the ones that status the attacker. I'm not sure if protect-bypassing moves also ignore that damage drop. That's another circumstance in which these moves would need to be considered.

Hyperspace Hole in particular also interacts similarly with Substitute, which is its own can of worms.

Moves affected up to current spreadsheet progress:
- Aerial Ace
- Phantom Force
- Shadow Force
- Smart Horn
- Magical Leaf
- Hyperspace Hole
- False Surrender
- Feint Attack
- Disarming Voice
- Acedian Impact
- Siege Breaker
- Swarm Order
- Shimmering Heat
- Blind Side
- Enchanting Song
- Pixie Punch

### LINKED ISSUE #149: Implement Moongeist Beam
Moongeist Beam applies a Mold Breaker effect. For that to be meaningful, abilities have to be implemented in the first place. We can get to it at the same time as we get to Juggernaut (nee Mold Breaker) itself, probably.

### LINKED ISSUE #156: Implement Brine
Dependent on the target's current HP. While this could be done with a custom input like with moves that scale based on the *user's* HP, there are other reasons to note the target's current HP, e.g. #19, so these systems should be unified.

### LINKED ISSUE #242: Implement "Overwhelm"
Requested on Discord

AlphaKretin:
Not expecting full testing of this, just a quick sanity check to make sure that pushing them won't blow up the website - we can let the users report if any of this doesn't work properly. Hoping to keep chipping away and adding more bit by bit over time so don't delete branch if merged, and may add more before review (or even after if I'm feeling really daring). 

Close #242 
Close #156 
Close #149 
Close #144 
Close #141 
Close #110 
Close #83


src/app/damagecalc/components/MoveCard.tsx
```diff
@@ -23,13 +23,18 @@ export interface MoveCardProps {
...
 export default function MoveCard(props: MoveCardProps): ReactNode {
-    const [crit, setCrit] = useState<boolean>(props.target.volatileStatusEffects.Jinx || props.moveData.criticalHit);
+    const forcesCrit =
+        props.moveData.move.forceCrit(props.user, props.target, props.battleState) ||
+        props.target.volatileStatusEffects.Jinx;
+    const [crit, setCrit] = useState<boolean>(forcesCrit || props.moveData.criticalHit);
     const [customInput, setCustomInput] = useState<unknown>(props.moveData.customVar);
 
+    // Sync moveData.criticalHit so the damage calc sees forced crits
+    props.moveData.criticalHit = crit || forcesCrit;
```
Cincidial: Is this making it so that the `crit` checkbox is checked when it's going to be a forced crit?

Not sure I'm just misunderstanding, I'm testing using talonflame with `harsh` against a burned target but not seeing it get checked automatically

src/app/data/abilities/OffensiveDamageBoostAbility.ts
```diff
@@ -0,0 +1,62 @@
...
+import { LoadedAbility } from "@/preload/loadedDataClasses";
+import { BattleState } from "../battleState";
+import { WeatherCondition } from "../conditions";
+import { Ability } from "../tectonic/Ability";
+import { PartyPokemon } from "../types/PartyPokemon";
+
+const sunWeathers: WeatherCondition[] = ["Sunshine", "Harsh Sunlight"];
+const rainWeathers: WeatherCondition[] = ["Rainstorm", "Heavy Rain"];
+const sandWeathers: WeatherCondition[] = ["Sandstorm"];
+const eclipseWeathers: WeatherCondition[] = ["Eclipse", "Ring Eclipse"];
```
Cincidial: Not sure if it's universally applicable, but could these be available as a global const somewhere else?

I see the DefensiveDamageAbility uses some that are similar, though I'm not sure if they could be reused

Cincidial (commented):
Couple nitpick style comments and a question. 

Also, I was thinking if it would be possible / if you've ever looked into a way to run the ruby battle code within JS. If it's possible it could save on the need to reimplement the battle logic and keep it up to date. I'm assuming such a thing is significantly easier said than done though

src/app/damagecalc/components/MoveCard.tsx
AlphaKretin: It's *supposed* to but if it's not working that's good to know

AlphaKretin:
> Also, I was thinking if it would be possible / if you've ever looked into a way to run the ruby battle code within JS. If it's possible it could save on the need to reimplement the battle logic and keep it up to date. I'm assuming such a thing is significantly easier said than done though

This is not something I've given a ton of thought, though that doesn't mean it's not *worth* thought. Ruby is portable enough that there's probably a way to run it in a web context, or maybe we can have an executable on the server that the website can call through an API...?

The reason I've been reimplementing stuff rather than copying Tectonic code directly is because there's a lot of stuff the Tectonic code does that isn't relevant to the damage calculator and that feels like unnecessary overhead. But maybe the extra overhead of reimplementation for us humans is something to be more concerned about. Also if we really were *directly* running the Tectonic code with no modification there's probably a bunch of shit that needs to be in scope for the ruby to run properly so it might wind up just looking like a headless version of the entire engine.

src/app/data/abilities/OffensiveDamageBoostAbility.ts
AlphaKretin: It seems very reasonable to have these be global yeah

src/app/damagecalc/damageCalc.ts
```diff
@@ -489,7 +502,7 @@ function pbCalcProtectionsDamageMultipliers(
 ): DamageMultipliers {
     // Aurora Veil, Reflect, Light Screen
     // TODO: Abilities that ignore screens?
```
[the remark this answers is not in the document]
AlphaKretin: The TODO certainly isn't now that it. Is done. Oops.
The header, despite the code below it making it pretty clear what it does, I think is still useful for skimming over code blocks without having to actually read each block

src/app/damagecalc/components/MoveCard.tsx
AlphaKretin: It looks like the reason Harsh didn't work is because force crit abilities aren't implemented lol, only moves. Testing Incision and Humble they seemed to work. I might implement Harsh later, but it's not a bug with this line at least.

Cincidial:
> I guess the first step in a proof of concept for this would be making a version of tectonic that strips out everything but battle code and accepts command line arguments to simulate damage. Then if that goes well we can look into hooking it up to the website somehow

Just as an aside - I've been working on some other stuff lately (hence not taking a look at the tectonic website a lot)
There is a way to run ruby code directly in the browser, you can use a WASM wrapper maintained by the community - https://github.com/ruby/ruby.wasm

I was able to get that up and running and actually calling ruby code (though nothing from the engine, just simple stuff like calling a function / printing to console). 

All in the way of saying though, there is a way to run the damage calc in the browser, it would just need be refactored to be functional instead of using so much global state lol

AlphaKretin:
Fascinating. That might be worth investigation at some point.

PR #380, opened 2026-02-17
Fix team code crash when importing ids that map to undefined

Cincidial:
Only affects the short form team code as what's happening here is the mapping is not finding the key provided and is returning undefined. 

Changed to use the `XYZ.NULL` value for the data, that way when the UI encounters it it just displays the default values for that data instead of crashing.

### DIFF
--- a/src/app/data/pokeparty.ts
+++ b/src/app/data/pokeparty.ts
@@ -254,8 +254,7 @@ export class PokePartyEncoding {
                         TectonicData.abilities,
                         Ability.NULL,
                     );
-                    mon.moves[0] =
-                        move1Id == 0 ? Move.NULL : (TectonicData.moves[decodingMapping.moves[move1Id]] ?? Move.NULL);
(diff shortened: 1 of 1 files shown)


src/app/data/pokeparty.ts
```diff
@@ -254,7 +254,8 @@ export class PokePartyEncoding {
                         TectonicData.abilities,
                         Ability.NULL,
                     );
-                    mon.moves[0] = move1Id == 0 ? Move.NULL : TectonicData.moves[decodingMapping.moves[move1Id]];
+                    mon.moves[0] =
+                        move1Id == 0 ? Move.NULL : (TectonicData.moves[decodingMapping.moves[move1Id]] ?? Move.NULL);
```
AlphaKretin: This seems redundant - if there's no move mapped to 0, wouldn't `TectonicData.moves[decodingMapping.moves[0]` be `undefined`, so `(TectonicData.moves[decodingMapping.moves[move1Id]] ?? Move.NULL` is all the code we need?

AlphaKretin (approved):

PR #381, opened 2026-02-21
Bugfix teambuilder decimal value

Petica98:
Seems like a relic from another commit, maybe the tracker/playthrough key couldnt have decimals before


Cincidial (approved):
Thanks for the fix!

AlphaKretin (approved):
Don't see any issue here. I was momentarily concerned about the keys being decimals - I thought maybe this fix was disguising a deeper bug - but Cinc's explanation makes it clear. Apologies for the wait on this, been swamped.

PR #402, opened 2026-05-24
Add optional battleState to calcTypeMatchup

### LINKED ISSUE #252: Gravity not nullifying ground immunities

AlphaKretin:
Allows for matchups to be conditional on battle elements, e.g. Gravity or abilities that check the weather
If battleStat isn't passed, it will be ignored, allowing the function to still be used for static type analysis like in the Pokedex
Fix #252

### DIFF
--- a/src/app/damagecalc/components/MoveCard.tsx
+++ b/src/app/damagecalc/components/MoveCard.tsx
@@ -104,7 +104,7 @@ export default function MoveCard(props: MoveCardProps): ReactNode {
                 className={`flex w-full justify-between items-center p-1 rounded-2xl ${getTypeColorClass(
                     props.moveData.move.getType(props.user, props.battleState),
                     "bg",
-                    "bg"
+                    "bg",
                 )} ${props.moveData.move.isSTAB(props.user) ? "font-bold" : ""} ${
                     props.moveData.move.isSignature ? "text-yellow-300" : ""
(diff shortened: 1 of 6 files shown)


src/app/damagecalc/components/MoveCard.tsx
```diff
@@ -165,7 +165,7 @@ export default function MoveCard(props: MoveCardProps): ReactNode {
                         props.battleState,
                         props.moveData.customVar
                     )}`}</span>
-                    <span>{`Accuracy: ${props.moveData.move.accuracy}%`}</span>
+                    <span>{`Accuracy: ${props.battleState.gravity && props.moveData.move.accuracy > 0 ? Math.min(props.moveData.move.accuracy * 2, 100) : props.moveData.move.accuracy}%`}</span>
```
Cincidial: Just curious, what move has 0% accuracy?

Cincidial (approved):

src/app/damagecalc/components/MoveCard.tsx
AlphaKretin: Moves that always hit, including status moves. On reflection though I'm not sure why we have to check that before doubling it. Really, instead, the 0% check should be elsewhere and not display accuracy at all in that case. Let me get that real quick before I merge this in.

PR #403, opened 2026-05-24
Add PBS Export to team builder

### LINKED ISSUE #352: Export trainer teams to PBS format
To allow devs to theorycraft in the builder and immediately export what they have to the game

AlphaKretin:
Dev only for now. When 3.4 releases we might make both this and Export to Tectonic public, but hidden behind an "advanced options" dropdown. 
Close #352


Cincidial (approved):
Don't see anything wrong in here. 
I'm assuming it remains dev only by not releasing to the live website? I don't see anything that would prevent it otherwise

AlphaKretin:
The export to PBS button is in the same block as the export to Tectonic button, guarded by an isDev check that didn't need to be added or changed. In any case, thanks for checking it over!

